PR #371, opened 2015-04-14
Update README with script from wiki

benzonico:
Hi, I just want to share my experience that running the script displayed on the readme on github did not work right away and using the script available on the wiki : https://aseba.wikidot.com/en:sourceinstall worked directly, hence the suggested update.

One more suggestion: how about making this into a script at the root of the project ?

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -35,21 +35,23 @@ Then you have to fetch Dashel, Enki, compile them, and then you can fetch Aseba 
 	mkdir -p aseba/build-dashel aseba/build-enki aseba/build-aseba
 	cd aseba
 	# fetch and compile dashel
-	git clone https://github.com/aseba-community/dashel.git
+	git clone git://github.com/aseba-community/dashel.git
 	cd build-dashel
-	cmake ../dashel -DCMAKE_BUILD_TYPE=RelWithDebInfo -DBUILD_SHARED_LIBS=OFF
+	cmake ../dashel -DCMAKE_BUILD_TYPE=RelWithDebInfo -DBUILD_SHARED_LIBS=ON
 	make
 	cd ..
 	# fetch and compile enki
-	git clone https://github.com/enki-community/enki.git
+	git clone git://github.com/enki-community/enki.git
 	cd build-enki
 	cmake ../enki -DCMAKE_BUILD_TYPE=RelWithDebInfo
 	make
 	cd ..
 	# fetch and compile aseba, telling it where to find dashel and enki
-	git clone https://github.com/aseba-community/aseba.git
+	git clone git://github.com/aseba-community/aseba.git
+	cd aseba && git checkout release-1.3.x && cd ..
 	cd build-aseba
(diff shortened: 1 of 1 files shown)


stephanemagnenat:
What is the error with the script from the README? From the diff of this pull request, the changes are:
- the use of git rather than https for accessing github, both should work
- the checkout of the release branch on the wiki, we do not want that in the readme
- the building of Dashel dynamically on the wiki and statically on the readme, that is on purpose as I find convenient to link development versions statically
- specifying twice dashel_DIR on the wiki, this seems useless but maybe there is an error in the syntax of the last line

benzonico:
Hi, 
A small disclaimer: I blindly copy pasted from the wiki for update and did not strip down to the real issue (and I am, by far no C++ and cmake expert, but I think this can be turned into valuable feedback on the documentation of the project).
I tried to build aseba yesterday on a machine at home. I agree that github https vs git is not the issue (both checkout were fine) nor checkout of the branch.

The issue I  faced seems to be linked to the static/dynamic building of Dashel. When statically built it could not find the correct cmake files when building aseba and failed.

I'll try to provide you more detailed logs of the encountered error with this README script if you are interested.

stephanemagnenat:
> I'll try to provide you more detailed logs of the encountered error with this README script if you are interested.

Yes please, it would be interesting to see what is the exact problem on your machine, maybe there is something to fix in our CMakeLists.txt files.

benzonico:
Ok I just reproduced that on another machine: 

OS: Ubuntu 14.04 64 bits
procedure : run the apt-get update command in Readme with no issue, copy paste the script provided in readme, run it and got the following output : 
```
CMake Error at examples/clients/CMakeLists.txt:8 (find_package):
  By not providing "Finddashel.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "dashel", but
  CMake did not find one.

  Could not find a package configuration file provided by "dashel" with any
  of the following names:

    dashelConfig.cmake
    dashel-config.cmake

  Add the installation prefix of "dashel" to CMAKE_PREFIX_PATH or set
  "dashel_DIR" to a directory containing one of the above files.  If "dashel"
  provides a separate development package or SDK, be sure it has been
  installed.


-- Configuring incomplete, errors occurred!
See also "/home/benzonico/Development/Thymio/aseba/build-aseba/CMakeFiles/CMakeOutput.log".
See also "/home/benzonico/Development/Thymio/aseba/build-aseba/CMakeFiles/CMakeError.log".
make: *** No targets specified and no makefile found.  Stop.
```

benzonico:
And in fact that seems to be it : adding the following line 
```
export dashel_DIR=../build-dashel
```
is enough to fix script on main page. 

I can update this PR to only add this relevant line if you wish.

stephanemagnenat:
I do not understand why `-Ddashel_DIR=../build-dashel` in the cmake command line does not work.

benzonico:
I am sorry to not be of anymore help about why the -D is not working. Pull request updated to end up with only the export of the env variable.

PR #382, opened 2015-05-26
asebahttp connects Aseba with HTTP

davidjsherman:
Provide a simple REST interface with introspection for Aseba devices.
 
    GET  /nodes                                 - JSON list of all known nodes
    GET  /nodes/:NODENAME                       - JSON attributes for :NODENAME
    PUT  /nodes/:NODENAME                       - install new Aesl program
    GET  /nodes/:NODENAME/:VARIABLE             - retrieve JSON value for :VARIABLE
    POST /nodes/:NODENAME/:VARIABLE             - send new values(s) for :VARIABLE
    POST /nodes/:NODENAME/:EVENT                - call an event :EVENT
    GET  /events[/:EVENT]*                      - create SSE stream for all known nodes
    GET  /nodes/:NODENAME/events[/:EVENT]*      - create SSE stream for :NODENAME
 
Typical use: `asebahttp -p 3000 -a vmcode.aesl ser:name=Thymio-II`

Variables and events are learned from the node description and parsed from AESL source when provided.
Server-side event (SSE) streams are updated as events arrive.
The two POST forms expect JSON, but also sloppily accept slash-delimited GET, for example 
* GET `/nodes/thymio-II/motor.left.target/100` sets the speed of the left motor
* GET `/nodes/thymio-II/sound_system/4/10` might record sound 4 for 10 seconds (if such an event were defined in AESL).

Compiled and tested on:
* MacOS Yosemite (clang-602.0.53)
* Ubuntu 12.04 (GNU g++ 4.6.3)
* Windows XP using MING32 under msys2-i686-20150202 (GNU g++ 4.9.2)

Unit tests (`tests/test-asebahttp`) require a dummy node at dashel target tcp:;port=33333.
They are defined using https://github.com/philsquared/Catch, which is included as a git submodule.

### DIFF
--- a/switches/http/README.md
+++ b/switches/http/README.md
@@ -1,0 +1,75 @@
+asebahttp - a switch to bridge Aseba to HTTP
+
+2015-01-01 David James Sherman <david dot sherman at inria dot fr>
+
+Provide a simple REST interface with introspection for Aseba devices.
+
+- GET  /nodes                                 - JSON list of all known nodes
+- GET  /nodes/:NODENAME                       - JSON attributes for :NODENAME
+- PUT  /nodes/:NODENAME                       - write new Aesl program (file= in multipart/form-data)
+- GET  /nodes/:NODENAME/:VARIABLE             - retrieve JSON value for :VARIABLE
+- POST /nodes/:NODENAME/:VARIABLE             - send new values(s) for :VARIABLE
+- POST /nodes/:NODENAME/:EVENT                - call an event :EVENT
+- GET  /events\[/:EVENT\]*                      - create SSE stream for all known nodes
+- GET  /nodes/:NODENAME/events\[/:EVENT\]*      - create SSE stream for :NODENAME
+
+Typical use: `asebahttp --port 3000 --aesl vmcode.aesl ser:name=Thymio-II &`
+After vmcode.aesl is compiled and uploaded, check with `curl http://127.0.0.1:3000/nodes/thymio-II`
+                
+Substitute appropriate values for :NODENAME, :VARIABLE, :EVENT and their parameters. In most cases,
+:NODENAME is `thymio-II`) For example.
+- `curl http://localhost:3000/nodes/thymio-II/motor.left.target/100`
+  sets the speed of the left motor
+- `curl http://nodes/thymio-II/sound_system/4/10`
+  might record sound number 4 for 10 seconds, if such an event were defined in AESL.
+                
+Variables and events are learned from the node description and parsed from AESL source when provided.
+Server-side event (SSE) streams are updated as events arrive.
+If a variable and an event have the same name, it is the EVENT that is called.
+On a local machine the server can handle 600 requests/sec with 10 concurrent connections,
+more (up to 2.5 times more) if the requests are pipelined as is the HTTP/1.1 default.
+
+Start an 'asebadummynode 0' and run 'make test' to execute some basic unit tests.
+Example [Node-RED](http://nodered.org) flows can be found in ../../examples/http/node-red.
+
+DONE:
+- Dashel connection to one Thymio-II and round-robin scheduling between Aseba and HTTP connections
+- read Aesl program at launch, upload to Thymio-II, and record interface for introspection
+- GET /nodes, GET /nodes/:NODENAME with introspection
+- POST /nodes/:NODENAME/:VARIABLE (sloppily allows GET /nodes/:NODENAME/:VARIABLE/:VALUE\[/:VALUE\]*)
+- POST /nodes/:NODENAME/:EVENT (sloppily allows GET /nodes/:NODENAME/:EVENT\[/:VALUE\]*)
+- form processing for updates and events (POST /.../:VARIABLE) and (POST /.../:EVENT)
+- handle asynchronous variable reporting (GET /nodes/:NODENAME/:VARIABLE)
+- JSON format for variable reporting (GET /nodes/:NODENAME/:VARIABLE)
+- implement SSE streams and event filtering (GET /events) and (GET /nodes/:NODENAME/events)
+- Aesl program bytecode upload (PUT /nodes/:NODENAME)
+  use curl --data-ascii "file=$(cat vmcode.aesl)" -X PUT http://127.0.0.1:3000/nodes/thymio-II
+- accept JSON payload rather than HTML form for updates and events (POST /.../:VARIABLE) and (POST /.../:EVENT)
+
+TODO:
+- handle more than just one Thymio-II node
+- gracefully shut down TCP/IP connections (half-close, wait, close)
+
+This code borrows from the rest of Aseba, especially switches/medulla and examples/clients/cpp-shell,
+which bear the copyright and LGPL 3 licensing notice below.
+
+
+/*
+Aseba - an event-based framework for distributed robot control
+Copyright (C) 2007--2015:
+Stephane Magnenat <stephane at magnenat dot net>
+(http://stephane.magnenat.net)
+and other contributors, see authors.txt for details
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU Lesser General Public License as published
+by the Free Software Foundation, version 3 of the License.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with this program. If not, see <http://www.gnu.org/licenses/>.
+*/
--- a/switches/http/http.cpp
+++ b/switches/http/http.cpp
@@ -1,0 +1,1263 @@
+
+/*
+ asebahttp - a switch to bridge Aseba to HTTP
+ 2014-12-01 David James Sherman <david dot sherman at inria dot fr>
+ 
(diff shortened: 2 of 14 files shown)


switches/http/do_valgrind_macosx.sh
```diff
@@ -0,0 +1,93 @@
+#!/bin/bash -x
```
stephanemagnenat: Why are these OS X only?

switches/http/example-nodered-simple.json
```diff
@@ -0,0 +1,126 @@
+[
```
stephanemagnenat: I would put examples codes in a subdirectory, and test code in another subdirectory. Alternatively, and maybe better, we should put example code in the top-level `example/` directory and test code in the top-level `tests/` directory. But I understand that because of dependencies, you might want to keep asebahttp stuff bundled together.

switches/http/README.md
```diff
@@ -0,0 +1,69 @@
...
+- gracefully shut down TCP/IP connections (half-close, wait, close)
+- remove dependency on libjson? Our output is trivial and input syntax is limited to unsigned vectors
+
+This code borrows from the rest of Aseba, especially switches/medulla and examples/clients/cpp-shell,
+which bear the copyright and LGPL 3 licensing notice below.
+
+
+/*
+Aseba - an event-based framework for distributed robot control
+Copyright (C) 2007--2012:
```
stephanemagnenat: I guess we should put 2015 instead of 2012 :-)

switches/http/http.cpp
```diff
@@ -0,0 +1,1280 @@
+
+/*
+ asebahttp - a switch to bridge HTTP to Aseba
+ 2014-12-01 David James Sherman <david dot sherman at inria dot fr>
+ 
+ Provide a simple REST interface with introspection for Aseba devices.
+ 
+    GET  /nodes                                 - JSON list of all known nodes
+    GET  /nodes/:NODENAME                       - JSON attributes for :NODENAME
```
stephanemagnenat: It is not clear to me whether `:NODENAME` means variable `NODENAME` or variable `:NODENAME`, and whether `:` should be put into the URL.

switches/http/http.cpp
```diff
@@ -0,0 +1,1280 @@
...
+    PUT  /nodes/:NODENAME                       - write new Aesl program (file= in multipart/form-data)
+    GET  /nodes/:NODENAME/:VARIABLE             - retrieve JSON value for :VARIABLE
+    POST /nodes/:NODENAME/:VARIABLE             - send new values(s) for :VARIABLE
+    POST /nodes/:NODENAME/:EVENT                - call an event :EVENT
+    GET  /events[/:EVENT]*                      - create SSE stream for all known nodes
+    GET  /nodes/:NODENAME/events[/:EVENT]*      - create SSE stream for :NODENAME
+ 
+ Server-side event (SSE) streams are updated as events arrive.
+ If a variable and an event have the same name, it is the EVENT the is called.
```
stephanemagnenat: grammar: /the EVENT the is called/

switches/http/http.cpp
```diff
@@ -0,0 +1,1280 @@
...
+#	error "You need at least Dashel version 1.0.3 to compile Http"
+#endif // DAHSEL_VERSION_INT
+
+//== Some utility functions and diverse hacks ==============================================
+
+// because I don't understand locales and wide strings yet
+char* wstringtocstr(std::wstring w, char* name)
+{
+    //char* name = (char*)malloc(128);
+    std::wcstombs(name, w.c_str(), 127);
```
stephanemagnenat: potential buffer overflow, as name might not be null-terminated if the maximum number of character is reached. Also, conversion error will lead to undefined data in `name`.

switches/http/http.cpp
```diff
@@ -0,0 +1,1280 @@
...
+            pendingResponses.erase(stream);
+            unsigned num = streamsToShutdown.erase(stream);
+            if (verbose)
+                cerr << stream << " Connection closed, removed " << num << " pending shutdowns" << endl;
+        }
+    }
+    
+    void HttpInterface::nodeDescriptionReceived(unsigned nodeId)
+    {
+        if (verbose)
```
stephanemagnenat: indentation

switches/http/http.cpp
```diff
@@ -0,0 +1,1280 @@
...
+                    appendResponse(subscriber->first, 200, true, reply.str().c_str());
+                }
+            }
+        }
+    }
+
+    //-- Routing for HTTP requests ---------------------------------------------------------
+
+    void HttpInterface::routeRequest(HttpRequest* req)
+    {
```
stephanemagnenat: Is `req` leaking? It is not deleted by the caller...

switches/http/http.cpp
```diff
@@ -0,0 +1,1280 @@
...
+                    VariablesMap vm = n->second;
+                    for (VariablesMap::iterator i = vm.begin();
+                         i != vm.end(); ++i)
+                    {
+                        json << (i == vm.begin() ? "" : ",")
+                             << "\"" << wstringtocstr(i->first, wbuf) << "\":" << i->second.second;
+                        seen_named_variables = true;
+                    }
+                }
+                if ( ! seen_named_variables )
```
stephanemagnenat: why extra spacing here?

switches/http/http.cpp
```diff
@@ -0,0 +1,1280 @@
...
+            }
+            size_t eventPos;
+            if (commonDefinitions.events.contains(UTF8ToWString("reset"), &eventPos))
+            {
+                strings data;
+                data.push_back("reset");
+                sendEvent(nodeName,data);
+            }
+            
+            this->unlock();
```
stephanemagnenat: `lock()` is commented...

switches/http/http.cpp
```diff
@@ -0,0 +1,1280 @@
...
+#include <libxml/tree.h>
+
+#if DASHEL_VERSION_INT < 10003
+#	error "You need at least Dashel version 1.0.3 to compile Http"
+#endif // DAHSEL_VERSION_INT
+
+//== Some utility functions and diverse hacks ==============================================
+
+// because I don't understand locales and wide strings yet
+char* wstringtocstr(std::wstring w, char* name)
```
stephanemagnenat: Functions for string conversions are provided in `common/utils/utils.h`, is there a reason not to use them?

switches/http/do_valgrind_macosx.sh
davidjsherman: This file can certainly be renamed. I think we can remove it, though, it has a hidden dependency on siege.

switches/http/example-nodered-simple.json
davidjsherman: Test code is already in top-level `tests`. I'll move the node-red example.

switches/http/http.cpp
davidjsherman: That's a common placeholder notation used in routing definitions (Ruby, Mojolicious). The other convention is {NODENAME}; would that be clearer? I want the reader to understand to show that pattern like `/nodes/:NODENAME/:EVENT/:PARAM1/:PARAM2` (or `/nodes/{NODENAME}/{EVENT}/{PARAM1}/{PARAM2}`) could be instantiated with a URL `http://localhost:3000/nodes/thymio-II/record_sound/4/60`.

switches/http/http.cpp
davidjsherman: No, it goes on pendingResponses and will be freed in the future by the incoming Aseba message (with the variable's value) that provides the response. Otherwise it is freed by unscheduleResponse when the stream shuts down.

switches/http/http.cpp
davidjsherman: Locking isn't working for me on Windows and ought to be removed here until it does.

stephanemagnenat: ok pour then `unlock()` should be commented as well?

switches/http/http.cpp
stephanemagnenat: Ok, I would add your response as a note on the creation of `req`.

switches/http/do_valgrind_macosx.sh
stephanemagnenat: Ok, then we should remove it.

switches/http/example-nodered-simple.json
stephanemagnenat: ok, good.

switches/http/http.cpp
stephanemagnenat: It is fine for me if it is a convention (I do not know the Ruby or Mojolicious ecosystem). Maybe we should just give an example to make clear how to pattern match it.

switches/http/valgrind-macosx.supp
```diff
@@ -0,0 +1,495 @@
+{
```
stephanemagnenat: Shouldn't this file go into the test directory as well?

switches/http/http.cpp
davidjsherman: No, the one-line bodies of the if(verbose) just don't have braces

stephanemagnenat: I meant in the following lines, there are 1 and 2 space indentation, while in the rest of the code it is 4 spaces.

switches/http/http.cpp
davidjsherman: comments about this promise on new and delete of HttpRequest

stephanemagnenat:
Looks good to merge now! Would you agree to do a rebase to have a single commit? That way, spurious changes to EventBlocks.cpp would be removed.

PR #393, opened 2015-08-07
Fixed qMin argument type for ARM

damieng002:
On ARM qReal is float and not double so we need to convert it first
This change should be compatible with all CPU architectures


stephanemagnenat:
Maybe then we should use std::min<double> instead...
-- 
http://stephane.magnenat.net

damieng002:
I try with std::min and it works on ARM and Intel architectures. I close this pull request and create a new one with std::min code

PR #403, opened 2015-08-19
asebahttp: verbose mode

derlin:
This update allows the user to launch asebahttp in verbose mode (-v option). 
To ease the debugging process, some repetitive output has been commented out.

### DIFF
--- a/switches/http/http.cpp
+++ b/switches/http/http.cpp
@@ -1057,13 +1057,13 @@ namespace Aseba
         for (StreamResponseQueueMap::iterator m = pendingResponses.begin();
              m != pendingResponses.end(); m++)
         {
-            if (verbose)
+            /*if (verbose)
             {
                 cerr << m->first << " sendAvailableResponses " << m->second.size() << " in queue";
                 for (ResponseQueue::iterator qi = m->second.begin(); qi != m->second.end(); qi++)
                     cerr << " " << *qi;
                 cerr << endl;
-            }
+            }*/
             bool close_this_stream = false;
             ResponseQueue* q = &(m->second);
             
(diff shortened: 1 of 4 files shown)


switches/http/http.cpp
```diff
@@ -1057,13 +1057,13 @@ namespace Aseba
         for (StreamResponseQueueMap::iterator m = pendingResponses.begin();
              m != pendingResponses.end(); m++)
         {
-            if (verbose)
+            /*if (verbose)
```
stephanemagnenat: why commenting this?

switches/http/http.cpp
```diff
@@ -1090,8 +1090,8 @@ namespace Aseba
                     q->pop_front();
                 }
             }
-            if (verbose)
-                cerr << m->first << " available responses sent, now " << m->second.size() << " in queue" << endl;
+            /*if (verbose)
```
stephanemagnenat: why commenting this?

derlin:
During my tests, I noticed that these lines are constantly printed at a rate that is kind of problematic. Indeed, other useful information is just hidden in their flow. Moreover, I always get the same line... I figured we could do without it, at least when we are in verbose mode (i.e. someone is probably watching the output in a console).

stephanemagnenat:
I see, @davidjsherman do you have an opinion on that?

targets/challenge/challenge-textures.qrc
```diff
@@ -13,6 +13,6 @@
 	<file>doc/challenge.fr.html</file>
 	<file>doc/challenge.es.html</file>
 	<file>doc/challenge.it.html</file>
-	<file>textures/asebachallenge.svgz</file
+	<file>textures/asebachallenge.svgz</file>
```
stephanemagnenat: I am surprised by this line, as it is already correct on the master branch...

stephanemagnenat:
A large refactoring of asebahttp took place recently, resulting in asebahttp2:
  https://github.com/aseba-community/aseba/tree/master/switches/http2
The latter should replace asbahttp and asebaswitch in the future. Asebahttp2 improved asebahttp along similar lines as your patches. Could you see if your patches are still relevant, and if so, rebase them against asebahttp2? Thank you!

PR #405, opened 2015-08-19
asebahttp: compilation errors

derlin:
By default, asebahttp returns `200: ok` to a `load` request whatever the outcome. However, in some cases the caller would be interested by the possible compilation errors. 

We modified the code so that the PUT data of a load request support a new argument: `verbose=true|false`. If set to true, the body of the response will contain compiler messages, or nothing if everything went well.

 Example:
-   `curl -X PUT --data-ascii "verbose=false,file=$(cat code.aesl)" http://localhost:3000/nodes/thymio-II/` behaves normally
-   `curl -X PUT --data-ascii "verbose=true,file=$(cat code.aesl)" http://localhost:3000/nodes/thymio-II/` returns an empty body upon  success and the error message (utf string) upon failure

Note that the verbose option should appear _before_ the file option for this to work.

### DIFF
--- a/switches/http/http.cpp
+++ b/switches/http/http.cpp
@@ -614,14 +614,24 @@ namespace Aseba
     
     void HttpInterface::evLoad(HttpRequest* req, strings& args)
     {
+        bool shouldReturnErrors = false;
+
         if (verbose)
             cerr << "PUT /nodes/" << args[0].c_str() << " trying to load aesl script\n";
         const char* buffer = req->content.c_str();
-        size_t pos = req->content.find("file=");
+
+        size_t pos = req->content.find("verbose=true");
         if (pos != std::string::npos)
         {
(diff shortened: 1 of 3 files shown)


stephanemagnenat:
@davidjsherman could you please review this patch? Thank you!

stephanemagnenat:
A large refactoring of asebahttp took place recently, resulting in asebahttp2:
  https://github.com/aseba-community/aseba/tree/master/switches/http2
The latter should replace asbahttp and asebaswitch in the future. Asebahttp2 improved asebahttp along similar lines as your patches. Could you see if your patches are still relevant, and if so, rebase them against asebahttp2? Thank you!

stephanemagnenat:
Because of new switch (#777), this PR is not valid any more.

PR #429, opened 2015-10-13
Fix building aseba-test-natives-count against static asebavm.

brettle:
The essence of the problem is that there is a circular dependency between asebavm and asebavmdummycallbacks, and they need to be linked into aseba-test-natives-count differently depending on whether asebavm is built as a shared lib (which it is when building RPMS but not when building packages for Ubuntu/Debian). This pull request fixes the problem by just using an "if (BUILD_SHARED_LIBS)" in the CMakeLists.txt file to control the 2 different cases.

NOTE: I don't have an Ubuntu/Debian system to test on. Someone else needs to confirm that this works there.

### DIFF
--- a/rpm/aseba.spec
+++ b/rpm/aseba.spec
@@ -9,7 +9,7 @@ Version:        %{source_major}.%{source_minor}.%{source_patch}
 
 # Update the following line with the git commit hash of the revision to use
 # for example by running git show-ref -s --tags RELEASE_TAG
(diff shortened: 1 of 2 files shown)


stephanemagnenat:
@FabianHahn, could you check if this PR solves your problem as well?

FabianHahn:
Yes, this fixes my linking problems and obsoletes #428.

PR #439, opened 2015-10-22
asebastudio: Add Thymio Blockly plugin

FabianHahn:
This PR adds a new asebastudio plugin that provides a Blockly interface for the Thymio, similar to what ThymioVPL provides. Most of the GUI code has been copied and adapted from ThymioVPL's main window.

The Blockly workspace is displayed within an embedded web browser provided by QtWebKit's `QWebView` widget. The web client is able to generate valid AESL code, which is then passed back to asebastudio and displayed in its editor. The `AsebaJavascriptInterface` class is used as an interface between asebastudio and the web application, and is implemented using the [Qt WebKit bridge](http://doc.qt.io/qt-4.8/qtwebkit-bridge.html) and the Qt signal system. Saving and loading is supported via Blockly's internal XML representation of a workspace, which can be embedded in an aesl file similar to how ThymioVPL stores its state. Rather than loading the HTML, JavaScript and media files for the web application from a web server, the Thymio Blockly fork is included as a Git submodule and the necessary sources are directly embedded into asebastudio as Qt resources.


stephanemagnenat:
Hi Fabian,

Thank you very much for the PR! One question: shouldn't we move your Blockly's tree into the aseba-community organisation?

FabianHahn:
Hi Stéphane,

Yes, I had also thought of that, I think that's a great idea - then we won't have to reference my personal fork here. Can you fork my repository from the aseba-community account, or ask whoever has control over it? I'd be happy to adjust the Git submodule in this PR once it is done.

Best,
Fabian

FabianHahn:
Changed to use aseba-community/blockly. Ready to merge from my side.

Issue #446, opened 2015-11-02
Behaviour for proximity and ground sensors blocks with grayed-out sensors


stephanemagnenat:
A user has reported that the current behaviour, triggering at 10 Hz, is intuitive for the proximity and ground sensors blocks. I agree, especially with the //when// behaviour introduced in 1.4. This situation needs to be re-thought for the next VPL release, the user suggests that the even never fires.

ddoak:
Do you mean the possible confusion caused by the "grayed-out sensors" block repeatedly triggering? e.g.
![vpl_sensor_beep_01](https://cloud.githubusercontent.com/assets/7423291/10879532/63aefa7a-814c-11e5-9bc4-63c7f5d71b9e.png)
will beep continously.

Also, a "grayed-out sensors" block can apparently mask other actions depending on the order of the VPL code. e.g. compare:
![vpl_sensor_colour_1a](https://cloud.githubusercontent.com/assets/7423291/10879815/4e28dc5a-814e-11e5-9c3f-74b4f121a211.png)
with:
![vpl_sensor_colour_1b](https://cloud.githubusercontent.com/assets/7423291/10879816/532c21d0-814e-11e5-9965-323da0c06a4f.png)
In the second case the red and green top led colour changes are never seen because they are immediately overwritten by a change to white.

stephanemagnenat:
Exactly :-)

cwalther:
(I was the user in question.) You probably meant to type *un*intuitive. :)

For the benefit of others reading, here’s the full description of my experience (which was shortly after seeing a Thymio and the VPL for the first time):

[...]

PR #452, opened 2015-11-12
Ignore proximity event blocks with grayed-out sensors

cwalther:
This implements the first part of what I suggested in #446, as a base for discussion: Proximity and button event blocks with all sensors grayed out should never trigger, rather than periodically, and cause a compiler warning.

It turns out that the current version of the VPL compiler is not capable of outputting warnings (while still generating code), only errors (which cause no code to be generated), so that’s what I implemented, as a first step. If there is agreement that that would be desirable, I can extend it to produce warnings instead. (MISSING_EVENT and MISSING_ACTION could be warnings too in my opinion, that would make it easier to test an incomplete program or “comment out” a line by removing its event or action.)

Is there anything I need to do to get the new UI strings internationalized? (I’m unfamiliar with Qt.)

If this is accepted, I will try to implement the second part: add a “periodic timer” block to replace the removed functionality.

### DIFF
--- a/clients/studio/plugins/ThymioVPL/Block.cpp
+++ b/clients/studio/plugins/ThymioVPL/Block.cpp
@@ -168,6 +168,11 @@ namespace Aseba { namespace ThymioVPL
 		return false;
 	}
 	
+	bool Block::needsAnyValueSet() const
+	{
+		return false;
+	}
+	
 	void Block::resetValues()
 	{
 		for (unsigned i=0; i<valuesCount(); ++i)
--- a/clients/studio/plugins/ThymioVPL/Block.h
+++ b/clients/studio/plugins/ThymioVPL/Block.h
@@ -83,6 +83,7 @@ namespace Aseba { namespace ThymioVPL
 		virtual void setValue(unsigned i, int value) = 0;
 		virtual QVector<quint16> getValuesCompressed() const = 0;
 		virtual bool isAnyValueSet() const;
+		virtual bool needsAnyValueSet() const;
 		void resetValues();
 		
 		virtual bool isAdvancedBlock() const { return false; }
--- a/clients/studio/plugins/ThymioVPL/Compiler.cpp
+++ b/clients/studio/plugins/ThymioVPL/Compiler.cpp
@@ -23,6 +23,7 @@
 #include <QObject>
 #include "Compiler.h"
 #include "Scene.h"
+#include "Block.h"
 
 namespace Aseba { namespace ThymioVPL
 {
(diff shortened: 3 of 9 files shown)


stephanemagnenat:
Thank you very much for this contribution. I am currently unsure whether to merge it or not. The reason is that any change to VPL requires to update a lot of documentation, and we are currently working on a new version that allows to implement state machines in an easier way. I will discuss with other Thymio actors and come back to you.

cwalther:
That’s fine, I don’t insist that it be merged as-is, it’s just a proposal for discussion.

Is any code or documentation or discussion available for the new VPL version? I’d be interested in following that. I checked the Aseba-dev mailing list but didn’t see anything there.

I am currently working on more changes to VPL: Adding a “periodic event” block that gives access to the Thymio’s second timer, restoring the accidental functionality that would be lost with this proposal and making things like blinking LEDs (the “hello world” of microcontroller programming) much simpler. The first version is working, but I need to clean up the code a bit before it can become a pull request. Also, while showing it to Dorit and other people at the FabLab last week, some ideas for improvement developed that I want to implement in a second variant. So this might have to be redesigned to work with that next version of VPL, or may be obsoleted by it altogether. But I don’t mind working on a potential dead-end for now, I’m doing this because I’m having fun doing it, not because I insist that the result is accepted.

stephanemagnenat:
The new VPL is being explored in this repository https://github.com/aseba-community/thymio-vpl2 with a preliminary bridge to Thymio in that one https://github.com/aseba-community/thymio-adventure, which will provide integration with augmented reality in the future, see https://github.com/aseba-community/thymio-ar

As you can see, all these are very preliminary and work in progress. The most advanced part is the "VPL2", which will certainly see its name change at some point.

cwalther:
Thanks, I’ll take a look. Unfortunately it seems like using Qt 5 on Mac OS X 10.6 is a hassle, if even possible at all, so it may take me a while.

stephanemagnenat:
I recently managed to compile the new VPL both on OS X and iOS. This new version will not be released without several rounds of testing. We are reaching an alpha version soon, at that point I will announce it to the mailing list. Help to test and report usability problems or improvements would be most welcome to. But feel free to try it already now and mail me if you encounter problems.

stephanemagnenat:
Note that VPL2 mobile preview is now [available on Google Play](https://play.google.com/store/apps/details?id=org.mobsya.thymiovpl2).

stephanemagnenat:
VPL2 mobile forbids grayed-out sensors. As VPL1 is not about to be updated as that would break current educational activities, I am closing this PR.

PR #454, opened 2015-11-14
Fix crash when saving a document that has not used ThymioBlockly

### LINKED ISSUE #449: crash when save file
In master branch Studio crashes when files are saved.
Test on my build and from new server

It works on  ThymioVPL stand alone

cwalther:
currentSavedXml in that case was an empty string, and passing that to document.setContent() apparently left the QDomDocument in an invalid state that would later crash MainWindow::saveFile().

Add some more error checking for good measure, although I am not currently aware of any circumstance where the error still occurs.

May fix #449.


stephanemagnenat:
Thank you!

cwalther:
Thanks, that was quick!

PR #469, opened 2016-01-09
Add a periodic timer event to VPL, variant 1

cwalther:
I propose adding a periodic timer event block to Thymio VPL, for the following reasons:

- It provides access to the Thymio’s second timer, which has previously been unavailable to VPL.

- It restores functionality that has previously existed by accident, but would be lost by #452. It does this in a better, more discoverable way and also makes it configurable.

- It makes it simpler to implement the “Hello World” of microcontroller programming, blinking an LED: This previously required using states to multiplex the functionality of the one available timer, as well as bootstrapping the process using another event – no more with the availability of both timers and the auto-starting behavior of the periodic timer.

This pull request is a prototype implementation of one variant. For a second variant, see pull request #470. Both are complete enough that they could be merged if deemed acceptable, but I mainly want to throw this out here as a basis for exploration and discussion. I do not insist that any of this be accepted at all, I just did it because I had fun doing it. A definitive integration would at least require additional updates to documentation.

This variant represents the periodic event using the metaphor of a heartbeat. Testing will have to tell whether users understand that. Here is the implementation of blinking with it:

blink-periodic1

I have seen Thymio-VPL2, but it seems sufficiently different from VPL1 that the two can coexist, and it does not seem to be in need of such a solution since it has other ways of creating periodic behavior, so I think it still makes sense to discuss this in the context of VPL1. Unless the decision is that VPL1 is feature-frozen now that development on VPL2 is underway.

What do you think, does the added functionality justify the added complexity?


stephanemagnenat:
Thank you very much Christian for this contribution. I have added this question to the agenda of our next internal Thymio meeting, and will come back to you after it.

stephanemagnenat:
Dear Christian. After discussing this question here at EPFL and with some of the educators using VPL, we decided not to add a new block for periodic timer in VPL1. The reason is that while a nice addition in absolute, it adds an option that is not totally necessary, because it can be simulated using the existing timer.

In VPL2 currently under development, we have not fixed the choice of blocks, and are therefore considering periodic timer as an interesting option.

kind regards! Stéphane

cwalther:
That’s fine, thanks for discussing it!

I have not kept up with VPL2 development, I probably should when I have time.

PR #470, opened 2016-01-09
Add a periodic timer event to VPL, variant 2

cwalther:
(continued from #469)

This variant uses the metaphor of a metronome to represent the periodic event and compared to variant 1 adds the possibility to trigger the event at either of two opposite phases of the period (on the “tick” or on the “tock” of the pendulum).

This addition arose from the insight that most of the actions available in VPL are idempotent, so that executing them periodically by themselves has no immediately observable effect. The effect only becomes apparent when between two executions another action is executed that undoes the first action, and it is often more convenient to trigger that second action from the same timer rather than requiring use of the other (non-periodic) timer.

While adding a small amount of UI complexity, this makes the implementation of blinking even simpler, it is now possible without using advanced mode:

blink-periodic2

On a technical note, I had to extend the VPL compiler a bit to implement this, it would be good to test it with some existing VPL files to ensure that it still generates the same code, but I didn’t have any such files available.


stephanemagnenat:
Thank you very much Christian for this contribution. I have added this question to the agenda of our next internal Thymio meeting, and will come back to you after it.

stephanemagnenat:
Dear Christian. After discussing this question here at EPFL and with some of the educators using VPL, we decided not to add a new block for periodic timer in VPL1. The reason is that while a nice addition in absolute, it adds an option that is not totally necessary, because it can be simulated using the existing timer.

In VPL2 currently under development, we have not fixed the choice of blocks, and are therefore considering periodic timer as an interesting option.

kind regards! Stéphane

PR #495, opened 2016-04-20
Silence various warnings on GCC 4.9.3

antoinealb:
When building latest Aseba master on ARM Cortex-M several warnings are produced. This PR fixes all of them and should not induce any behaviour changes.

### DIFF
--- a/transport/can/can-buffer.c
+++ b/transport/can/can-buffer.c
@@ -23,12 +23,14 @@
 
 void AsebaSendBuffer(AsebaVMState *vm, const uint8* data, uint16 length)
 {
+	(void) vm;
 	while (AsebaCanSend(data, length) == 0)
 		AsebaIdle();
 }
 
 uint16 AsebaGetBuffer(AsebaVMState *vm, uint8* data, uint16 maxLength, uint16* source)
 {
+	(void) vm;
 	return AsebaCanRecv(data, maxLength, source);
 }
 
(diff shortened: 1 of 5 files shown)


transport/can/can-buffer.c
```diff
@@ -23,12 +23,14 @@
 
 void AsebaSendBuffer(AsebaVMState *vm, const uint8* data, uint16 length)
 {
+	(void) vm;
```
stephanemagnenat: I guess this aims at removing a "variable not used" warning. I was not aware about this usage before this pull request, do you have a link to a reference/rationale for it? I'm curious!

antoinealb: You are correct: this prevents warnings about unused variables. I do not have a "serious" reference available but I have seen it used on several open source projects. See also this SO answer: http://stackoverflow.com/a/308286

stephanemagnenat:
I have added an ASEBA_UNUSED macro in commit 4b4cbc522f8da83269ad20e606665a10ce5f05a3, could you please update the pull request to use it? It is more informative than using (void) directly.

antoinealb:
Sure. I rebased my PR and added `ASEBA_UNUSED`.

transport/can/can-buffer.c
```diff
@@ -20,15 +20,18 @@
 
 #include "../buffer/vm-buffer.h"
 #include "can-net.h"
+#include "common/consts.h"
```
stephanemagnenat: This path should be relative, "../../common/consts.h"

antoinealb: Ok, wasn't aware of that. Fixed now.

PR #498, opened 2016-04-22
Support for easy out-of-source build

tomas789:
Hi,

I've added support for cmake's out of source build. I added dashel, enki and qwt into thirdparty subdirectory and compile them directly from aseba's CMakeLists.txt. This simplifies building process a lot. I also rewrote parts of readme.md to describe this. 

Because there is no "standard aseba way" to test software I did only some testing. I tested build on OS X 10.11 and Windows 8 and it worked well. Everythink was compiled using few straight-forward cmake-ish way.

I beleave this can deliver a bit of user-friendlyness in build process :)

I'm looking forward to see this pull request in aseba repository. If you have any questions feel free to comment this request. I'll answer as fast as I can.

Thank you for Aseba!


stephanemagnenat:
Thank you for this contribution!

I need to discuss internally to see what is our policy regarding the build of Aseba. It is not trivial, because we have a build server already set up so changing this could lead to significant work.

Regarding your pull request, I have some comments:

- In the main `CMakeLists.txt` you force the link to `libdashel.a`, which forces the creation of static library for Dashel.
- You provided some changes to Aseba Challenge which are not relevant to this build-related pull request. If you wish to make changes to Aseba Challenge, please provide a separate request.
- You imported the whole Qwt source tree, this is not clean. It would be better to use submodules like for Dashel and enki. Some people already provide a github repository for Qwt, see for instance https://github.com/osakared/qwt.

Thank you very much!

tomas789:
Hello,

I hope simplicity of build process will overweight additional effort with changing build process. Do you use some public CI server like travis-ci? It would be useful for the community to make it public.

To respond your's comments:

- Could you describe why is this not good?
- I'm sorry. This is a mess from my other modifications to for school project.
- I didn't found Qwt's official git repository so I considered this to be more "clean" then relying at unofficial repositories.

stephanemagnenat:
> Do you use some public CI server like travis-ci?

We use our internal jenkins running on a Mac computer, with Ubuntu and Windows virtual machines inside virtualbox guests.

> It would be useful for the community to make it public.

I agree, note that all our build scripts are public, see https://github.com/aseba-community/build-server, https://github.com/aseba-community/aseba-windows and https://github.com/aseba-community/aseba-osx.

> I hope simplicity of build process will overweight additional effort with changing build process.

That's true, but we have to estimate the cost of changing our build system now, as we are short of resources. I'm travelling now, I'll discuss this with my colleagues this week.

> Could you describe why is this not good?

This leads to larger executables than strictly necessary, but it's not a drama I concede.

> I'm sorry. This is a mess from my other modifications to for school project.

If we decide to merge your pull request, I will ask you to clean this before.

> I didn't found Qwt's official git repository so I considered this to be more "clean" then relying at unofficial repositories.

I see your point, but I partially disagree. A temporary solution would be to have our own import of Qwt in the `aseba-community` organization, but if the unofficial repository is clean, I'm fine with using it.

davidjsherman:
The build system ([asebacommunity/aseba-windows](https://github.com/aseba-community/aseba-windows), [asebacommunity/aseba-osx](https://github.com/aseba-community/aseba-osx)) already resolves this issue. I do agree that the build system might not be entirely user-friendly, at least not presently, but I believe that Tomas's improvements should be incorporated over there and not in Aseba itself. 

To my mind Dashel, Enki, Qt, Qwt are shared libraries that ought to be installed at the system level because they can be used in other software. They shouldn't be submodules of Aseba. This is different from [Thymio Blockly](https://github.com/aseba-community/blockly), which is tightly coupled to Aseba and belongs in the source code tree, but which is a fork of another package and needs to be kept in its own repository.

(Catch shouldn't be a submodule either, it's just one header and standard practice is just to [download it on demand with CMake](https://github.com/philsquared/Catch/blob/master/docs/build-systems.md#cmake).)

The use of submodules in the build system also makes sense because it may be desirable to pin the precise commit for the release.

PR #501, opened 2016-05-02
Ignore start address record in hexfiles

antoinealb:
arm-none-eabi-objcopy emits that type of record to indicate program entry point. I think it should be ignored because otherwise it rejects perfectly valid files.

### DIFF
--- a/common/utils/HexFile.cpp
+++ b/common/utils/HexFile.cpp
@@ -219,6 +219,26 @@ namespace Aseba
 						throw WrongCheckSum(lineCounter, checkSum, computedCheckSum);
 				}
 				break;
+
+				case 5: // start linear address record
+				{
+					// Start linear address record are not used by the Aseba
+					// bootloader protocol so we can ignore them.
+
+					// Skip data
+					for (int i = 0; i < 4; i++)
+						computedCheckSum += getUint8(ifs);
+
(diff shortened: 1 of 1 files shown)


common/utils/HexFile.cpp
```diff
@@ -219,6 +219,28 @@ namespace Aseba
...
 				}
 				break;
+
+				case 5: // start linear address record
+				{
+					// Start linear address record are not used by the Aseba
+					// bootloader protocol so we can ignore them.
+
+					// Skip data
+					for (int i = 0; i < 4; i++) {
```
stephanemagnenat: Please put `{` on its own line to respect Aseba's coding style.

common/utils/HexFile.cpp
```diff
@@ -219,6 +219,28 @@ namespace Aseba
...
+						computedCheckSum += getUint8(ifs);
+					}
+
+					computedCheckSum = 1 + ~computedCheckSum;
+
+					uint8 checkSum = getUint8(ifs);
+
+					if (checkSum != computedCheckSum) {
```
stephanemagnenat: Please put `{` on its own line to respect Aseba's coding style, or in this case you can omit the `{`, `}` pair.

antoinealb:
Fixed style as requested.

PR #502, opened 2016-05-03
asebahttp acts as a switch between multiple Aseba targets

davidjsherman:
Highlights of this pull request:
- maintain a set of Aseba streams and forward messages between them
- listen to a tcpin: target to offer an Aseba target for other clients
- use Aseba protocol 5 and ping the network every 1 second
- perform end-to-end testing with asebahttp, asebaswitch, asebadummynode and asebamassloader

The end-to-end tests comprise 456 assertions in 110 tests and are described in [tests/e2e-http/README.md](https://github.com/davidjsherman/aseba/blob/master/tests/e2e-http/README.md)


davidjsherman:
Note that for clarity this pull request does **not** include changes to asebascratch, since they are still highly experimental. They can be found in my downstream fork, most recently in branch use-R_state davidjsherman/aseba@fabdced.

switches/http/http.cpp
```diff
@@ -134,65 +135,89 @@ namespace Aseba
...
-    
     void HttpInterface::run()
     {
         do
         {
             sendAvailableResponses();
-			pingNetwork();
-            if (!run2s())
-				break;
+            step(2);
```
stephanemagnenat: The replacement of `run2s()` by `step(2)` is not correct, as step will wait /at most/ 2 seconds, and might return early if data arrive.

davidjsherman: I admit that I missed adapting this from 31d63747853e5ed9c5c3e215102730c887d7afec. Testing with `if (!run1s()) break;` (line 275, but see my comment line 306)

switches/http/http.cpp
```diff
@@ -252,84 +271,221 @@ namespace Aseba
...
+                    if (verbose)
+                        cerr << stream_to_shutdown << " shutting down stream" << endl;
+                    shutdownStream(stream_to_shutdown);
+                }
+                catch(Dashel::DashelException& e)
+                { }
+            }
+
+            // standard Aseba run loop
+            if (!step(timeout))
```
davidjsherman: Using `timeout` seems to throttle communications, I get better results with a low value like `50`.

stephanemagnenat: I am surprise, as on POSIX platforms `step()` calls `poll()`, see aseba-community/dashel@a1f385a780e4252e39e59157e4db0c5b7d28fc29/dashel/dashel-posix.cpp line 1179. On Windows, it is implemented with a `WaitForMultipleObjects` call.

Is this effect happening with all platforms?

switches/http/http.cpp
stephanemagnenat: > Testing with `if (!run1s()) break;`

Any result?

switches/http/http.cpp
davidjsherman: After more testing I've concluded that it doesn't affect throughput: both `sleep(timeout)` and `sleep(50)` can handle 40 transactions/second with 5 concurrent HTTP connections (thus 200/sec overall), with no failed transactions. This is true for asebahttp and asebascratch.

The behavior is easy to see in Scratch using asebascratch, but might be an interaction with Scratch, so I created an issue davidjsherman/aseba#24.

PR #523, opened 2016-05-26
Playground walls can have angles and mass, plus an example maze

davidjsherman:
Wall elements in playgrounds are defined by a width, length, and height around a center. This PR adds  an ‘angle’ attribute to wall elements in order to specify a rotation around the center, expressed in radians.

As an illustration, `examples/thymio-maze-6x6.playground` contains a 6x6-cell maze with two Thymios, that can be used to test navigation strategies based on wall avoidance or on track following.

Here is a sample run with one [Investigator](https://www.thymio.org/en:thymiobehaviourinvestigator) robot and one [Explorer](https://www.thymio.org/en:thymiobehaviourexplorer) robot:
example-maze-6x6-2


targets/playground/playground.cpp
```diff
@@ -211,6 +211,10 @@ int main(int argc, char *argv[])
 			wallE.attribute("h").toDouble(),
 			-1
 		);
+        if (! wallE.attribute("angle").isNull())
+        {
+            wall->angle = wallE.attribute("angle").toDouble(); // radians
+        }
```
stephanemagnenat: please indent with tabs within existing code, thank you ;-)

davidjsherman:
Indentation is fixed.
Wall and cylinders can also have finite mass, which allows them to be pushed.
To illustrate, I added a playground for the R2T2 (Remote Rescue using Thymio2) activity.

run-r2t2

stephanemagnenat:
Thank you very much, it is a pleasure to receive your contributions :-)

davidjsherman:
The pleasure is mine!
The complete R2T2 scene would need something for the generator in the middle. It could be simulated with two E-pucks running around the middle pushing a red block, looking for the blinks from the Thymios with their cameras. I think this could be done if the LED color of the Thymios is made available as a surface texture.

ddoak:
This looks great!
Looking forward to playing with it.

Issue #544, opened 2016-08-20
asebadummynode should allow ephemeral ports (and advertise using Zeroconf)


davidjsherman:
Asebadummynode takes an optional argument 0–9 for its node type, and hardwires a TCP port in the range 33333–33342. This behavior is expected by many programs.

We should allow dynamic allocation of ephemeral ports as described in [RFC 6335 Internet Assigned Numbers Authority (IANA) Procedures for the Management of the Service Name and Transport Protocol Port Number Registry](https://tools.ietf.org/html/rfc6335#section-6). Since `bind` will do this in Posix and Windows if the requested port is 0, it should be enough to use `port=0` in the Dashel target.

I propose adding an optional parameter to asebadummynode:
- --port|-p INTEGER|any

to allow asebadummynode to request any port, with a value of 'any' or '0' indicating that a dynamic port number should be chosen from the ephemeral range. Note that this will also allow creating multiple instances of dummynode-_i_ with the same node id on the same host machine.

Since the caller will not know what port was allocated, asebadummynode could print the Dashel target, or at least the port number. Ideally, asebadummynode would advertise itself using Zeroconf as described in issue #543 “Advertise targets using Zeroconf”.

stephanemagnenat:
I agree. Also, the switch should have a way to list this port on HTTP. One way could be to list all Aseba targets plus the listening port. In JSON it is something like:
``{ asebaTargets = [ "ser:name=Thymio-II", "tcp:localhost;33335" ], asebaPort = 33333 }``

davidjsherman:
It looks like this is slightly tricky. A Dashel `tcpin:` stream will indeed receive a dynamically allocated ephemeral port if its target parameters contain `port=0`, but Dashel doesn't check what port was assigned. The way to find the port number is by calling `getsockname` (available in posix and winsock) on the file descriptor. This would require a modification to Dashel: we can't call `getsockname` ourselves, since the file descriptor is protected in Dashel::SelectableStream.

In Dashel, the constructors for SocketServerStream and UDPSocketStream could call `getsockname` after `bind` and record the assigned port in the parameter set.

A dynamic port number can't be reserved in advance, so we can't just reserve one and provide it to Dashel.

stephanemagnenat:
I've implemented the feature implemented in Dashel commit aseba-community/dashel@0686c134a48c858cc4aff86bb7c9c4827f0c9b75

davidjsherman:
[...]

PR #547, opened 2016-08-25
asebadummynode can request ephemeral ports

davidjsherman:
First part of issue #544: asebadummynode now can ask for an ephemeral port using the `--port 0` option. Additionally, the `--basePort` option allows one to start numbering anywhere, not just 33333.
```
% asebadummynode -p 0
tcp:port=59048
% asebadummynode -p 20000 1
tcp:port=20000
% asebadummynode -b 20000 2
tcp:port=20002
% asebadummynode
tcp:port=33333
```
This PR depends on Dashel >= 1.2.0 in order to correctly print the target with the generated port number; otherwise, it is reported as `tcp:port=0`.

```
Usage: asebadummynode [--basePort|-b BASE, default 33333] [ID, from 0 to 9]
Usage: asebadummynode --port|-p PORT [ID, from 0 to 9]
Usage: asebadummynode --help|-h
Creates one node dummynode-ID with node id ID+1 listening on port:
 - BASE+ID, if BASE != 0 and BASE+ID is available
 - PORT, if PORT != 0 and PORT is available
 - a dynamically chosen port, if PORT or BASE == 0.
The Dashel target is printed on stdout.
```

### DIFF
--- a/targets/dummy/dummynode.cpp
+++ b/targets/dummy/dummynode.cpp
@@ -313,13 +313,12 @@ extern "C" void AsebaAssert(AsebaVMState *vm, AsebaAssertReason reason)
 
 int usage(char* program)
 {
-	std::cerr << "Usage: " << program << " [--basePort|-b BASE, default 33333] [ID, from 0 to 9]" << std::endl;
-	std::cerr << "Usage: " << program << " --port|-p PORT [ID, from 0 to 9]" << std::endl;
+	std::cerr << "Usage: " << program << " [--port|-p PORT] [ID, from 0 to 9]" << std::endl;
 	std::cerr << "Usage: " << program << " --help|-h" << std::endl;
 	std::cerr << "Creates one node dummynode-ID with node id ID+1 listening on port:" << std::endl;
-	std::cerr << " - BASE+ID, if BASE != 0 and BASE+ID is available" << std::endl;
+	std::cerr << " - a dynamically chosen port, if PORT == 0" << std::endl;
 	std::cerr << " - PORT, if PORT != 0 and PORT is available" << std::endl;
-	std::cerr << " - a dynamically chosen port, if PORT or BASE == 0." << std::endl;
+	std::cerr << " - 33333+ID, if PORT is not set and 33333+ID is available." << std::endl;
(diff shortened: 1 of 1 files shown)


stephanemagnenat:
Thank you for these features. I am wondering if BASE is necessary. Why not simply `-p PORT [ID]`?

davidjsherman:
If we are certain that the legacy auto-increment behavior is only to avoid port conflicts, then there is no particular reason. I don't know whether there is a scenario in use out there where predictable port numbers for sequential node ids are assumed.

I admit that there is not much difference between the two lines
```
for i in 0 1 2 3 4 5 6 7 8; do asebadummynode -p $[ 20000 + $i ] $i & done
for i in 0 1 2 3 4 5 6 7 8; do asebadummynode -p 20000 $i & done
```

davidjsherman:
Getting the targets without Zeroconf is not easy because one needs to read stdout of what is typically a backgrounded process. One way to do it is:
```
targets=$( for i in 0 1 2; do exec 3< <(asebadummynode -p 0 $i); read <&3; echo $REPLY; done; )
```
using [process substitution](https://www.gnu.org/software/bash/manual/bash.html#Process-Substitution) in Bash.

stephanemagnenat:
The legacy auto-increment behavior is currently only to avoid port conflicts, as the new switch does automatically remap conflicting ids. The original switch did not, so auto-incrementing the node ID was a good solution. Therefore, I suggest to keep it simple and only use -p PORT [ID] as command-line argument.

I think that either setting port by hand, or using zeroconf, or using process substitution covers a large enough range of cases.

targets/dummy/dummynode.cpp
```diff
@@ -82,29 +82,32 @@ class AsebaNode: public Dashel::Hub
 		vm.variablesSize = sizeof(variables) / sizeof(sint16);
 	}
 	
-	void listen(int basePort, int deltaPort)
+	Dashel::Stream* listen(const int port, int deltaNodeId)
```
stephanemagnenat: Why port is const and not deltaNodeId?

stephanemagnenat:
Perfect, let's merge, thank you :-)

PR #551, opened 2016-09-03
register apidocs with each operation

davidjsherman:
The apidocs for each operation comprise at least seven fields, including a map that defines every response code the operation can produce. Since this information is contained the JSON-formatted OAS specification for the REST API, it is easiest to provide the apidocs in JSON format.

These fragments are easy to extract using `jq`; for example, POST /constants is
```
jq -c '.paths."/constants".post' aseba-v1.json
```

This PR adds an `apidocs` map parallel to `handlers`, and a variant `registerHandler` that accepts apidocs. It also modifies `getApidocs` to provide global OAS information about the API.

(TODO: include model definitions, for $ref schema references.)

### DIFF
--- a/switches/http2/http/HttpDispatcher.cpp
+++ b/switches/http2/http/HttpDispatcher.cpp
@@ -258,12 +258,14 @@ namespace Aseba
 	{
 		handlers[method][uriPath] = handler;
 
+		// Since opId wasn't supplied by apidocs, construct an opId from the method and the URI Template path elements.
+		// For example, endpoint GET /nodes/{node}/variables has default opId GET-nodes-node-variables.
 		std::stringstream opId;
 		opId << toString(method);
-		for (auto elt: uriPath) // copy elt so can remove braces
+		for (auto element: uriPath) // copy path element so can remove braces
 		{
-			elt.erase(std::remove_if(elt.begin(), elt.end(), [](char c){ return c=='}'||c=='{'; }), elt.end());
-			opId << "-" << elt;
+			element.erase(std::remove_if(element.begin(), element.end(), [](char c){ return c=='}'||c=='{'; }), element.end());
+			opId << "-" << element;
 		}
 		apidocs[method][uriPath] = json{ {"operationId", opId.str()} };
 	}
@@ -276,7 +278,7 @@ namespace Aseba
 
 	void HttpDispatcher::optionsHandler(HandlerContext& context)
 	{
-		HttpResponse response(HttpResponse::fromStatus());
+		HttpResponse response{HttpResponse::fromJSON(buildApiDocs())};
 		response.headers.emplace("Access-Control-Allow-Methods", "GET, PUT, POST, DELETE, OPTIONS");
 		response.headers.emplace("Access-Control-Allow-Headers", "Content-Type");
 		response.send(context.stream);
@@ -289,7 +291,13 @@ namespace Aseba
 	
 	void HttpDispatcher::getApiDocs(HandlerContext& context)
 	{
-		json response = R"({"swagger":"2.0","schemes":["http"],"host":"localhost:3000","info":{"version":"1","title":"Aseba","description":"REST API for Aseba"},"parameters":{"trait:serverSentEventStream:todo":{"name":"todo","in":"query","required":false,"type":"integer"}},"responses":{"trait:serverSentEventStream:200":{"description":"stream of server-sent events","schema":{"type":"string"}}}})"_json;
+		json response{buildApiDocs()};
+		HttpResponse::fromJSON(response).send(context.stream);
+	}
+
+	json HttpDispatcher::buildApiDocs()
+	{
+		json documentation = R"({"swagger":"2.0","schemes":["http"],"host":"localhost:3000","info":{"version":"1","title":"Aseba","description":"REST API for Aseba"},"parameters":{"trait:serverSentEventStream:todo":{"name":"todo","in":"query","required":false,"type":"integer"}},"responses":{"trait:serverSentEventStream:200":{"description":"stream of server-sent events","schema":{"type":"string"}}}})"_json;
 		for (const auto& handlerMapKV: handlers)
 		{
 			for (const auto& handlerKV: handlerMapKV.second)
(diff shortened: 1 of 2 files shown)


davidjsherman:
It would be better to assign default documentation in `registerHandler`, rather than in `getApidocs`.

switches/http2/http/HttpDispatcher.h
```diff
@@ -98,6 +99,9 @@ namespace Aseba
 		typedef std::map<HttpMethod, URIHandlerMap> HandlersMap;
 		//! Handlers for known method + URI couples
 		HandlersMap handlers;
+		//! Map of method, split URI to JSON-formatted OAS documentation
+		typedef std::map<HttpMethod, std::map<strings, json>> ApidocsMap;
```
stephanemagnenat: There is one entry of this map for one entry in `handlers`, right? If so I would build a `pair<Handler, json>` as key in handlers.

switches/http2/http/HttpDispatcher.cpp
```diff
@@ -51,7 +51,8 @@ namespace Aseba
 		REGISTER_HANDLER(getApiDocs, GET, { "apidocs" });
 		REGISTER_HANDLER(getConstantsHandler, GET, { "constants" });
 		REGISTER_HANDLER(putConstantsHandler, PUT, { "constants" });
-		REGISTER_HANDLER(postConstantsHandler, POST, { "constants" });
+		REGISTER_HANDLER(postConstantsHandler, POST, { "constants" },
+						 R"({"tags":["Definitions – Constants"],"summary":"Create Constant Definition","description":"","operationId":"POST-constant-definition","responses":{"201":{"description":"","schema":{"$ref":"#/definitions/constant-definition-full"}},"400":{"description":"","schema":{"type":"string"},"examples":{"application/json":"Invalid value"}},"403":{"description":"","schema":{"type":"string"},"examples":{"application/json":"Invalid constant name"}},"409":{"description":"","schema":{"type":"string"},"examples":{"application/json":"Constant already exists"}}},"consumes":["application/json"],"produces":["application/json"],"parameters":[{"name":"body","in":"body","schema":{"$ref":"#/definitions/constant-definition-partial"}}]})"_json);
```
stephanemagnenat: Really cool the user defined literal, is it defined in the JSON module we use?

switches/http2/http/HttpDispatcher.cpp
```diff
@@ -256,8 +257,23 @@ namespace Aseba
 	void HttpDispatcher::registerHandler(const Handler& handler, const HttpMethod& method, const strings& uriPath)
 	{
 		handlers[method][uriPath] = handler;
+
+		std::stringstream opId;
```
stephanemagnenat: I do not understand what this code does

switches/http2/http/HttpDispatcher.cpp
```diff
@@ -273,15 +289,25 @@ namespace Aseba
...
+				for (auto elt: handlerKV.first) // copy so can modify
+					uri << "/" << elt;
+				// method key must be lower case for OAS 2.0
+				auto method = toString(handlerMapKV.first);
+				std::transform(method.begin(), method.end(), method.begin(), ::tolower);
+
+				// apidoc for this operation should have been registered
+				json doc = apidocs[handlerMapKV.first][handlerKV.first];
+				// patch this operation into the apidocs
+				response["paths"][uri.str()][method] = doc;
```
stephanemagnenat: Why do we have to patch something here?

switches/http2/http/HttpDispatcher.cpp
davidjsherman: Yes! https://github.com/nlohmann/json#serialization--deserialization

switches/http2/http/HttpDispatcher.cpp
davidjsherman: It removes the braces from the path element in the URI Template. I admit it is ugly.
For example the default opId for GET /nodes/{node}/variables is `GET-nodes-node-variables`, the same as the one that would be generated by the Stoplight API designer.

switches/http2/http/HttpDispatcher.cpp
stephanemagnenat: I just see that it is added to the response, so this is ok, I first thought the API docs structure was patched.

switches/http2/http/HttpDispatcher.cpp
stephanemagnenat: Ok, so we should just add a small comment line explaining that. Also it took me a while to get that `elt` meant `element`.

switches/http2/http/HttpDispatcher.cpp
```diff
@@ -273,15 +289,25 @@ namespace Aseba
...
 			{
-				response.push_back({
-					{ "method", toString(handlerMapKV.first) },
-					{ "uri", handlerKV.first }
-				});
+				if (handlerKV.first.size() == 0)
+					break;
+				// build uri string with slash delimiters
+				std::stringstream uri;
+				for (auto elt: handlerKV.first) // copy so can modify
```
stephanemagnenat: You can use the function `join<strings>("/")` from utils.h for that.

switches/http2/http/HttpDispatcher.cpp
davidjsherman: The complete api-docs object needs to be constructed from a basic JSON object, with general information about the API (keys `host`, `info`, `schemes`, `swagger`), model definitions (key `definitions`), and then an entry under key `path` for every endpoint.

So I first create the JSON object (line 292, in this version without the `definitions`), then insert the endpoint in line 310. The comment says "patch" because I was thinking of JSON Patch ([RFC 6902](https://tools.ietf.org/html/rfc6902)), but in fact it is just an insertion.

switches/http2/http/HttpDispatcher.h
davidjsherman: Assuming that nobody every uses the same handler with more than one method or wants to register uriPath aliases (registerHandler allows this)...

switches/http2/http/HttpDispatcher.cpp
davidjsherman: Fixed in fedde91f934e6ade95fa371c8110c3e09e6459f7

switches/http2/http/HttpDispatcher.cpp
davidjsherman: Clarified in fedde91f934e6ade95fa371c8110c3e09e6459f7 line 317

davidjsherman: using join() from utils.h in fbdebc0e2eab970118c3679ff9a4cec2f09f427e

switches/http2/http/HttpDispatcher.h
stephanemagnenat: One can use the same handler for more than one method, as the method+path is the key. I just propose to use both the handler and its documentation as value. The same handler for a different path would have a different documentation I guess. Currently there are two exactly parallel maps.

PR #554, opened 2016-09-11
Aseba::Zeroconf provides Zeroconf (mDNS-SD) services

davidjsherman:
Using Zeroconf, an Aseba target announces itself on the TCP/IP network as a service of type `_aseba._tcp`, and can be discovered by clients. A target might be an individual node or an Aseba switch. The zeroconf service provides minimal information about a target: Dashel tcp: target (host name, IP port), target name, target type, Aseba protocol version, a list of node ids, and a list of node product ids. A client may obtain complete node descriptions by connecting to the Dashel target and requesting them using the Aseba protocol.

`Aseba::Zeroconf` is a smart container for `Aseba::Zeroconf::Target` and can be iterated.

`Aseba::Zeroconf::Target` records information about a TCP Dashel target, which can be advertised by its owner or can be discovered by DNS Service Discovery.

`Aseba::Zeroconf::TxtRecord` provides creation and serialization of DNS TXT records for targets.

`Aseba::Zeroconf::ZeroconfDiscoveryRequest` is a private class that encapsulates service requests.


common/CMakeLists.txt
```diff
@@ -6,6 +6,18 @@ set (ASEBACOMMON_SRC
 	msg/msg.cpp
 	msg/NodesManager.cpp
 )
+
+find_path(DNSSD_INCLUDE_DIR dns_sd.h)
```
stephanemagnenat: There is a missing search for the library. I understand that on OS X it is linked by default, but checking the header without specifying the library will likely break compilation on Linux and Windows.

Ah, if only we had a CI system that would try to merge and compile all pull requests ;-)

common/zeroconf/target.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+using namespace std;
+
+namespace Aseba
+{
+	//! This target is described by a human-readable name and a port.
+	//! If a port is nonzero at creation then this is a local target to be remembered
+	//! otherwise it is a remote one discovered through browsing and may be refreshed.
+	//! Note that later the port will become nonzero if the target is resolved.
+	Zeroconf::Target::Target(const std::string & name, const int & port) :
```
stephanemagnenat: It is a detail but I do not see why a int (4 bytes) should be passed as reference (8 bytes)

common/zeroconf/txtrecord.cpp
```diff
@@ -0,0 +1,118 @@
...
+			fields[key] = ""; // length will be 0, so no "=" in the record
+	}
+
+	//! TXT data for an Aseba target, deserialized from a TXT record buffer
+	Zeroconf::TxtRecord::TxtRecord(const unsigned char * txtRecord, uint16_t txtLen)
+	{
+		size_t pos = 0;
+		while (pos < txtLen)
+		{
+			auto length{size_t(txtRecord[pos++])};
```
stephanemagnenat: Strange syntax, wouldn't `size_t length(txtRecord[pos++])` work?

common/zeroconf/txtrecord.cpp
```diff
@@ -0,0 +1,118 @@
...
+			string data{reinterpret_cast<const char*>(txtRecord)+pos, min(length,txtLen-pos)};
```
stephanemagnenat: Also strange syntax for me, why not `string data(reinterpret_cast<const char*>(txtRecord)+pos, min(length,txtLen-pos));` which is more explicit.

common/zeroconf/txtrecord.cpp
```diff
@@ -0,0 +1,118 @@
...
+	//! TXT data for an Aseba target, deserialized from a TXT record string
+	Zeroconf::TxtRecord::TxtRecord(const std::string & txtRecord)
+	{
+		istringstream txt{txtRecord};
+		do
+		{
+			char length;
+			txt >> length;
+			string data;
+			txt >> std::setw(int(length)) >> data;
```
stephanemagnenat: We are using namespace std, is there a reason to specify it again?

common/zeroconf/txtrecord.cpp
```diff
@@ -0,0 +1,118 @@
...
+		serialize_field(txt, "type");
+		serialize_field(txt, "ids");
+		serialize_field(txt, "pids");
+		return txt.str();
+	}
+	void Zeroconf::TxtRecord::serialize_field(ostringstream& txt, const std::string& key) const
+	{
+		if (fields.find(key) != fields.end()) // key will be absent if it was boolean and false
+		{
+			string record{key};
```
stephanemagnenat: Same comment as earlier, why not using `()`?

common/zeroconf/zeroconf.h
```diff
@@ -0,0 +1,210 @@
...
+		//! Retrieve the encoded TXT record
+		virtual std::string record() const;
+
+		virtual const std::string at (const std::string& k) const { return fields.at(k); }
+		virtual std::map<std::string, std::string>::iterator begin() { return fields.begin(); }
+		virtual std::map<std::string, std::string>::iterator end() { return fields.end(); }
+
+	private:
+		std::map<std::string, std::string> fields;
+		virtual void serialize_field(std::ostringstream& txt, const std::string& key) const; //!< serialize a field into the encoded TXT record
```
stephanemagnenat: Why not `serializeField`?

common/zeroconf/zeroconf.h
```diff
@@ -0,0 +1,210 @@
...
+	 \addtogroup zeroconf
+
+		A Zeroconf::Target allows client classes to choose and access Aseba targets.
+	 */
+	class Zeroconf::Target
+	{
+	public:
+		std::string name;
+		int port;
+		std::string domain{"local."};
```
stephanemagnenat: We could also use the `=` syntax, as it would match the initialization of static fields in C++98, but `{}` is also fine, as it denotes a C++11 syntax. Just we should write this choice somewhere so that we are consistent.

davidjsherman:
According to build job https://ci.inria.fr/inirobot/job/zeroconf/4/, commit 18626329d7b8d2747307030af9fe15484ecf3e60 compiles on macOS and Ubuntu trusty with GCC 4.8.4

common/zeroconf/txtrecord.cpp
davidjsherman: Trying to consistently use C++11 uniform initializers.

stephanemagnenat:
Ok, if it is good for gcc it is good for me :-) I thought `netinet/in.h` was not available in Windows, but maybe it is.

common/zeroconf/txtrecord.cpp
stephanemagnenat: Ok, I find it look excessively complicated for simple types such as integers and strings, but it's a matter of taste.

davidjsherman:
No you are right, for windows we would need Winsock1.h

common/CMakeLists.txt
davidjsherman: Would this work?
https://github.com/Eyescale/CMake/blob/master/FindDNSSD.cmake

davidjsherman:
Commit 78f7c96cdbdbd2845b489b317f543415e22e2abb compiles on 
* macOS 10.9.5 (AppleClang 6.0.0.6000051)
* Ubuntu trusty (GCC 4.8.4)
* Windows 7 (GCC 5.3.0).

https://ci.inria.fr/inirobot/job/zeroconf/7/
https://ci.inria.fr/inirobot/job/zeroconf_windows/3/

davidjsherman:
Seems to work on Windows 7 and macOS.
The compatibility layer for Avahi on Ubuntu throws an exception for DNSServiceBrowse

common/zeroconf/zeroconf.h
```diff
@@ -0,0 +1,211 @@
...
+#include <typeinfo>
+#include <algorithm>
+#include <string>
+#include <iostream>
+#include <sstream>
+#include <iomanip>
+#include <map>
+#include <vector>
+#include <thread>
+#include <dns_sd.h>
```
stephanemagnenat: As the path for this file is included explicitly in the CMakeLists.txt, shouldn't we use `#include "dns_sd.h"`, in case there is a conflicting system-wide version?

davidjsherman:
The code compiles and links on all three OSs. Simple tests using #555 work on Windows 7 and macOS. Browsing and resolving hang on Ubuntu, because the dns_sd compatibility layer in Avahi blocks in `DNSServiceProcessResult`. The fix for this is probably to implement the event loops in #556.

I think this PR can be merged.

Issue #233, opened 2013-09-02
Thresholds for sensors


stephanemagnenat:
In advanced mode, it would be nice to be able to alter the sensor thresholds.

riedo:
I just noticed something while doing tests (btw, the "grey" state of the sensors helps a lot with the hands following, and reordering is not problematic anymore):

The sliders map values linearly between 0 and 4000, but in reality, when nothing is detected the value is 0, and as an object approaches it suddenly "jumps" to more than 1500; in addition, it often reaches more than 5000 if the hand comes really close. It depends on the sensors and thus it is not easy to find universal values values; however, in the current situation, only looking at VPL the position of the slider does not help much to predict at which distance the robot will react. 
For a hand following (just back and forth, without touching the hand) I could not do it with just VPL; I had to look at the text and auto-refreshed memory window to understand the problem. I add a screenshot underneath to explain the problem; it's not very intuitive looking IMO.

![screen shot 2015-08-10 at 3 07 00 pm](https://cloud.githubusercontent.com/assets/1862778/9172174/eef80364-3f71-11e5-8631-0924d5f447ad.png)
Should we consider doing some measures on several robots and adapt the value mapping on the sliders?

motib:
Hi Fanny,
I have complained of the variation in the horizontal prox sensors and I'm sure many others have too. Stephane send me data from a large number of trials, and the values are widely distributed (my robot had even more extreme values), so it is hopeless to fix thresholds that are good for everyone.

I don't understand why you have to observe the absolute values in Studio. In advanced mode the thresholds can be adjusted by trial-and-error -- a well-known problem-solving technique :-)). The problem only exists in basic mode with its fixed thresholds and I don't think it's too hard to go to advanced mode if someone is having a serious problem in basic mode. That's why I don't complain about this issue any longer!

Please explain what the second line of your program does. The Thymio follows my hand easily when I remove the line.
Moti

PS: Stephane, why is the issue still open? Your original post asked for adjustable threshold and they have been implemented.

riedo:
Maybe I'm just not patient enough with the trial and error :)
I looked at the numerical values because the logic of my code seemed good, but it did not work, and by changing a bit the thresholds I could not really measure the progress. With the feedback from the robot it's easier to compare and know which threshold is too high or too low.

[...]

PR #556, opened 2016-09-11
Aseba::Zeroconf derived classes for asynchronous responses

davidjsherman:
The basic `Aseba::Zeroconf` service works synchronously, and will block while waiting for responses from the DNS Service Discovery dæmon. These derived classes implement different mechanisms for detecting asynchronous updates and triggering methods in client classes. Three examples in `examples/` illustrate these mechanisms.

`Aseba::DashelhubZeroconf` proposes a Dashel::Hub that listens to DNS Service Discovery file descriptors using ~~Dashel::SocketStreams~~ Dashel::PollSocketStreams (introduced in aseba-community/dashel#19) and calls _xxxCompleted()_ methods that can be overridden in derived classes.

`Aseba::ThreadZeroconf` uses a std::thread to listen to DNS Service Discovery file descriptors using _poll(2)_ and calls _xxxCompleted()_ methods that can be overridden in derived classes.

`Aseba::QtZeroconf` uses QSocketNotifiers within the QT event loop to listen to DNS Service Discovery file descriptors and emit QT signals to which QObject slots may be connected.


stephanemagnenat:
Although I merged PR 555 into master, this PR still shows part of the content already in PR 555 in _Files Changed_.

common/zeroconf/zeroconf-dashelhub.h
```diff
@@ -0,0 +1,41 @@
...
+	protected:
+		//! Set up function called after a discovery request has been made. The file
+		//! descriptor associated with zdr.serviceref must be watched, to know when to
+		//! call DNSServiceProcessResult, which in turn calls the callback that was
+		//! registered with the discovery request.
+		virtual void processDiscoveryRequest(ZeroconfDiscoveryRequest & zdr);
+
+	private:
+		//! Collection of (tcp:) SocketStreams that watch the serviceref file descriptors.
+		std::vector<Dashel::Stream *> zeroconfStreams;
```
stephanemagnenat: Do we need these streams, cannot we reuse the ones already in the `Dashel::Hub`?

davidjsherman: The streams provided by the base `Dashel::Hub` may be any streams. We need to remember which ones we created to listen to `ZeroconfDiscoveryRequest` sockets, so that in `incomingData` we can call `DNSServiceProcessResult` on the right servicerefs.

So should really be `std::map<Dashel::Stream *, ZeroconfDiscoveryRequest &>`

stephanemagnenat: i agree for the map, there is a similar one in the switch.

davidjsherman:
Draft implementation of `Aseba::DashelhubZeroconf` using `DashelHub`, with modified example `asebatargetlist` to test the Hub API. Every pending service request's file descriptor *N* becomes a Dashel `tcp:sock=N` stream

Problem: `Dashel::step` never wakes up from a poll of these sockets :expressionless:

common/zeroconf/zeroconf-dashelhub.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+namespace Aseba
+{
+	using namespace Dashel;
+
+	void DashelhubZeroconf::processDiscoveryRequest(ZeroconfDiscoveryRequest & zdr)
+	{
+		if (int socket = DNSServiceRefSockFD(zdr.serviceref) != -1)
+		{
+			string dashel_target = FormatableString("tcp:sock=%0").arg(socket);
```
stephanemagnenat: Why not `dashelTarget`?

stephanemagnenat:
I have a vague souvenir you told me that there was a main socket connected to the zeroconf daemon that was used to notify the client that there was data to process. If my memory is correct, shouldn't this socket be read first?

davidjsherman:
No, there is a socket for **each** pending request. Using *select(2)* the socket needs to be in the read FD_SET; `ThreadZeroconf::handleDnsServiceEvents` in zeroconf-thread.cpp does the right thing. Dashel uses *poll(2)* with `POLLIN` set and it seems to me that that should be the same.

From the overview in the [DNS Service Discovery Programming Guide](https://developer.apple.com/library/content/documentation/Networking/Conceptual/dns_discovery_api/Articles/registering.html):

> If the registration can be started, DNSServiceRegister initializes the service discovery reference and creates a socket that is used to communicate with the mDNSResponder daemon. Use the service discovery reference to call DNSServiceRefSockFD and get the socket descriptor for the service reference.
> 
> Set up a run or select loop using the socket descriptor. When the loop indicates that the mDNSResponder daemon’s reply is available, call DNSServiceProcessResult and pass to it the service discovery reference initialized by DNSServiceRegister. DNSServiceProcessResult will call the callback function associated with the service discovery reference.

stephanemagnenat:
Yes, _poll(2)_ should do the same as _select(2)_. Maybe some debugging with _strace_ or an equivalent program is necessary?

davidjsherman:
Yes, *poll_emu* uses *select(2)* and seems to do the right thing. Since ce2de415a29e5fc6f6d3dd5fac4499fcee07c6e4 Dashel listens on the socket returned by `DNSServiceRefSockFD` and triggers `incomingData` so that `DNSServiceProcessResult` can be called on the service reference.
**But**, now `DNSServiceProcessResult` is hanging on those service references. 🙄

davidjsherman:
As of 6bbf05269f209b48c2e8c93de2d6470a5bef621b the implementation of `ThreadZeroconf` and `QtZeroconf` work, insofar as the simple tests in the examples `asebatargetlist-thread` and `asebatargetlist-qt` can tell.

However `DashelhubZeroconf` still hangs in `DNSServiceProcessResult`.

examples/zeroconf/CMakeLists.txt
```diff
@@ -4,14 +4,45 @@ add_executable(asebatargetadvertise
...
+)
+target_link_libraries(asebatargetlist-thread ${ASEBA_CORE_LIBRARIES} ${DNSSD_LIBRARIES})
+
+
+# Using QtZeroconf
+
+if (QT4_FOUND)
+        set(QT_USE_QTNETWORK ON)
+	set(QT_DONT_USE_QTGUI ON)
+        set(CMAKE_AUTOMOC ON)
```
stephanemagnenat: Do we still need automoc here?

examples/zeroconf/targetlist-dashelhub.cpp
```diff
@@ -0,0 +1,69 @@
...
+			// Aseba::Zeroconf is a smart container for Aseba::Zeroconf::Target
+			for (auto & target: targets)
+				// Resolve the host name and port of this target, retrieve TXT record
+				target.resolve();
+		}
+
+		void resolveCompleted(const Aseba::Zeroconf::Target * target)
+		{
+			// output could be JSON but for now is Dashel target [Target name (DNS domain)]
+			std::cout << target->host << ";port=" << target->port;
```
stephanemagnenat: Why using `std::` prefix as we are in the cpp file and `using namespace std;`.

examples/zeroconf/targetlist-qt.cpp
```diff
@@ -0,0 +1,84 @@
...
+				todo.insert(&target);
+				// Resolve the host name and port of this target, retrieve TXT record
+				target.resolve();
+			}
+		}
+
+		void resolveCompleted(const Aseba::Zeroconf::Target * target)
+		{
+			// output could be JSON but for now is Dashel target [Target name (DNS domain)]
+			std::cout << target->host << ";port=" << target->port;
```
stephanemagnenat: Why using `std:: prefix` as we are in the cpp file and using `namespace std;`.

examples/zeroconf/targetlist-thread.cpp
```diff
@@ -0,0 +1,86 @@
...
+				todo.insert(&target);
+				// Resolve the host name and port of this target, retrieve TXT record
+				target.resolve();
+			}
+		}
+
+		void resolveCompleted(const Aseba::Zeroconf::Target * target)
+		{
+			// output could be JSON but for now is Dashel target [Target name (DNS domain)]
+			std::cout << target->host << ";port=" << target->port;
```
stephanemagnenat: Why using `std::` prefix as we are in the cpp file and using `namespace std;`.

stephanemagnenat:
Do you want to merge now, or do you prefer to fix `DashelhubZeroconf` first?

davidjsherman:
We can merge, the code compiles on the three platforms [osx](https://ci.inria.fr/inirobot/job/zeroconf/label=inirobot-osx/lastSuccessfulBuild/) [ubuntu](https://ci.inria.fr/inirobot/job/zeroconf/label=inirobot-u64/lastSuccessfulBuild/) [win7](https://ci.inria.fr/inirobot/job/zeroconf_windows/label=inirobot-win7/lastSuccessfulBuild/). Even though there is a known bug and there are not yet any unit tests.

But, I would like your opinion on `QtZeroconf`, whose interface seems clumsy. Currently the class signals events using the same derived-class approach as the others. I don't program Qt, but to my unlearnèd eye it seems to me that it would be more natural to declare public SIGNALs to connect to client slots. The code signaling currently in the `asebatargetlist-qt` example could migrate to the `QtZeroconf` class.

davidjsherman:
The problem with `DashelhubZeroconf` is definitely that Dashel reads from the service reference socket, and consumes the data that `DNSServiceProcessResult` is expecting. There doesn't seem to be a way in the current Dashel to ask it to poll a socket without reading.

It might be possible to define an *ad hoc* `PollSocketStream : Dashel::SocketStream` whose `receiveDataAndCheckDisconnection` method doesn't read any data.

common/zeroconf/zeroconf-qt.h
```diff
@@ -0,0 +1,86 @@
...
+	class QtZeroconf : public QObject, public Zeroconf
+	{
+		Q_OBJECT
+
+	public:
+		void browse();
+
+	signals:
+		void zeroconfBrowseCompleted(); //!< emitted when browsing is completed
+		void zeroconfRegisterCompleted(const Aseba::Zeroconf::Target *); //!< emitted when a register is completed
```
stephanemagnenat: This signal passes a pointer to the target, however this signal might be connected to another thread through a `Qt::QueuedConnection`. In that case, are we sure that the target pointer will still be valid at that time. I guess this pointer points to a member of `targets` in the main `Zeroconf` object.

If this is a problem, we could split the current `Zeroconf::Target` object into a parent that contains all the informations but the request itself and the pointer to the container, and a child that adds these information. Then, a copy of the parent would be passed to the callee slot.

common/zeroconf/zeroconf-dashelhub.h
```diff
@@ -0,0 +1,54 @@
...
+
+namespace Aseba
+{
+	//! Use the Dashel run loop to watch the DNS Service for updates, and trigger callback
+	//! processing as necessary.
+	class DashelhubZeroconf : public Zeroconf, public Dashel::Hub
+	{
+	public:
+		void browse();
+		void run();
```
stephanemagnenat: I do not think this method and `run2s()` should be in this object, because the typical use case of this class would be with streams other than zeroconf-ones, therefore I think it is up to the client to implement the running policy. Moreover, for this specific method, it just forwards to the Hub one so it is useless.

common/zeroconf/zeroconf-thread.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+	void ThreadZeroconf::run()
+	{
+		watcher.join();
+	}
+
+
+	void ThreadZeroconf::processDiscoveryRequest(ZeroconfDiscoveryRequest & zdr)
+	{
+		zeroconfDRs.insert(&zdr);
```
stephanemagnenat: Missing lock when accessing this `zeroconfDRs`, which is shared between two threads.

common/zeroconf/zeroconf-thread.cpp
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	void ThreadZeroconf::eraseDiscoveryRequest(ZeroconfDiscoveryRequest & zdr)
+	{
+		zeroconfDRs.erase(zeroconfDRs.find(&zdr));
```
stephanemagnenat: Missing lock when accessing this `zeroconfDRs`, which is shared between two threads.

common/zeroconf/zeroconf-thread.cpp
```diff
@@ -0,0 +1,100 @@
...
+				continue;
+			}
+			fd_set fds;
+			int max_fds(0);
+			FD_ZERO(&fds);
+			std::map<DNSServiceRef,int> serviceFd;
+
+			int fd_count(0);
+
+			for (auto const& zdr: zeroconfDRs)
```
stephanemagnenat: Missing lock when accessing this `zeroconfDRs`, which is shared between two threads.

davidjsherman:
Current status in db87dacb65913a2d795bbde1e56ffd4e50e1ba8e:

platform | Compile | base | threads | dashelhub | qt
---- | ---- | ---- | ---- | ---- | ----
macOS | ✔️ | ✔️ | ✔️ | ✔️ | ✔️
Win7 | ✔️ | ✔️ | segfault | -65551 | ✔️
Ubuntu | ✔️ | hangs | -65544 | -65544 | -65544

Error -65544 (kDNSServiceErr_Unsupported) is a problem with the Avahi compatibility layer.
Error -65551 (kDNSServiceErr_Incompatible) is “client library incompatible with daemon”

common/zeroconf/zeroconf-thread.cpp
```diff
@@ -0,0 +1,100 @@
...
+		zeroconfDRs.erase(zeroconfDRs.find(&zdr));
+	}
+
+	void ThreadZeroconf::handleDnsServiceEvents()
+	{
+		struct timeval tv{1,0}; //!< maximum time to learn about a new service (5 sec)
+
+		while (running)
+		{
+			if (zeroconfDRs.size() == 0)
```
stephanemagnenat: Missing lock when accessing this `zeroconfDRs`, which is shared between two threads.

common/zeroconf/zeroconf-thread.cpp
```diff
@@ -0,0 +1,100 @@
...
+					max_fds = max_fds > fd ? max_fds : fd;
+					FD_SET(fd, &fds);
+					serviceFd[zdr->serviceref] = fd;
+					fd_count++;
+				}
+			}
+			int result = select(max_fds+1, &fds, (fd_set*)NULL, (fd_set*)NULL, &tv);
+			if (result > 0)
+			{
+				for (auto const& zdr: zeroconfDRs)
```
stephanemagnenat: Missing lock when accessing this `zeroconfDRs`, which is shared between two threads.

common/zeroconf/zeroconf-thread.cpp
```diff
@@ -0,0 +1,100 @@
...
+					if (FD_ISSET(serviceFd[zdr->serviceref], &fds))
+						DNSServiceProcessResult(zdr->serviceref);
```
stephanemagnenat: Is it safe to call this from the watcher thread?

common/zeroconf/zeroconf-thread.cpp
```diff
@@ -0,0 +1,100 @@
...
+			}
+			else if (result < 0)
+				throw Zeroconf::Error(FormatableString("handleDnsServiceEvents: select returned %0 errno %1").arg(result).arg(errno));
```
stephanemagnenat: The exception will be thrown from the watcher thread, I'm afraid it won't reach the handler which will likely lie in thread which created the `ThreadZeroconf` object.

davidjsherman:
Now in a16b457437f6576cfecad0bc8e7f4996b0fcde39:

1. ThreadZeroconf locking with `std::lock_guard`
1. ThreadZeroconf exception sharing with `std::exception_ptr`
1. Various cleanup as per comments

Thanks these changes and to updates in aseba-community/dashel#19, specifically that Dashel::PollStream in Win32 doesn't consume data from the socket, the current status is:

platform | Compile | base | threads | dashelhub | qt
---- | ---- | ---- | ---- | ---- | ----
macOS | ✔️ | ✔️ | ✔️ | ✔️ | ✔️
Win7 | ✔️ | ✔️ | ✔️ | ✔️ | ✔️

davidjsherman:
The problem seen with Ubuntu is due to the Avahi compatibility layer, which only partially implements the Bonjour API.

1. Avahi doesn't ever set `kDNSServiceFlagsMoreComing` (lathiat/avahi#18 is over a year old and we can't expect users to update Avahi soon), so some browse results arrive after `Zeroconf::browseCompleted` has been called; a flag can fix this but a timeout will also be necessary.
1. The poll file descriptors are being called twice for the same event, once with POLL_DONE and once with POLL, and Avahi's `avahi_simple_poll_dispatch` is calling `Hub::incomingData` twice. There may be something in the Avahi API to avoid this.

If this can't be made to work, it will be necessary to write an Avahi native implementation of `Aseba::Zeroconf`.

stephanemagnenat:
The compatibility layer is on the client side, right? If so, could we use a patched copy locally while still communicating with the distribution-provided server?

davidjsherman:
Yes, I think we can use a patched copy of the compatibility layer in asebacommon and still link to the Avahi library provided by the distribution. This custom compatibility layer will communicate with the client side API of the Avahi library.

davidjsherman:
1. Patched Avahi compatibility layer 53af3732fde070cb18d8cc8c9427868a0d29133d
1. Exception for out of order browse responses 1371475db634a01b20e5aec4c897989ef77c9fb0
1. PollStream doesn't prematurely close socket davidjsherman/dashel@82a309258502c34b7eab18bde742d873df529753

platform | Compile | base | threads | dashelhub | qt
---- | ---- | ---- | ---- | ---- | ----
macOS | ✔️ | ✔️ | ✔️ | ✔️ | ✔️
Win7 | ✔️ | ✔️ | ✔️ | ✔️ | ✔️
Ubuntu | ✔️ | hangs | ✔️ | ✔️ | ✔️

davidjsherman:
So, the synchronous behavior used in the base class doesn't work with Avahi. There is an old ticket from 2011 [#233: Bonjour compatibility layer - Synchronous DNSServiceResolve behaviour not present](http://mail.0pointer.net/pipermail/avahi-tickets/2011-September/001532.html) that is still not addressed.

Also, I discovered that the Avahi compatibility layer starts a thread for every service ref anyway.

davidjsherman:
Since this is a bug in a system library, and we should never encounter it because every client we have will use one of the derived classes for asynchronous polling, I am inclined to flag this **WON'T FIX**. I was not able to correct the bug in Avahi nor find a workaround.

I think that this PR can be merged.

stephanemagnenat:
I agree for the merge, after one small question is fixed: in the Qt version, I believe that there can be a problem with the memory in case of queued connexion as `const Aseba::Zeroconf::Target * target` might point to an already-deleted target. I put a comment but you did not address it.

davidjsherman:
I see your comment, let me check

davidjsherman:
An `Aseba::Zeroconf` is a container for `Aseba::Zeroconf::Target`. It contains local targets that are registered to it, and a cache of remote targets that have been discovered by mDNS-SD. Targets are stored in a member `std::vector<Target> targets` and the STL container operations on `Aseba::Zeroconf` are provided by that `std::vector`.

A `Target` sent in another thread by a `Qt::QueuedConnection` may no longer exist, either because it is a local target that has been erased from the container or because it is a remote target that is no longer in the cache: the cache is cleared at the start of every `browse` operation. The other thread should not keep the `Target` object alive in either of these cases, but should gracefully handle the disappearance of the object.

It seems to me that it should be possible to wrap the `Target` in a `QWeakPointer` in the signals.

Changing the container to `std::vector<std::shared_ptr<Target>> targets` would complicate all of the non-QT code and I would like to avoid that if I can.

stephanemagnenat:
I still suggest the same solution as in my original comment:

> If this is a problem, we could split the current Zeroconf::Target object into a parent that contains all the informations but the request itself and the pointer to the container, and a child that adds these information. Then, a copy of the parent would be passed to the callee slot.

davidjsherman:
Unless I have misunderstood it, your suggestion is still an encapsulation that will cascade changes through the existing code; in that case isn't it just as easy to use `std::vector<std::shared_ptr<Target>>`?

davidjsherman:
Maybe I can limit the encapsulation to just the Qt derived class. Am I right that a Qt application will have one thread with Zeroconf, and a GUI thread that needs to be informed about updates?

stephanemagnenat:
> Unless I have misunderstood it, your suggestion is still an encapsulation that will cascade changes through the existing code;

Yes my  suggest is to modify existing code, but it is a light change, and it won't affect any user code beside examples as none has been written yet. I can do the change in a branch wrt master if you want, then you can decide if you like it.

> Maybe I can limit the encapsulation to just the Qt derived class. Am I right that a Qt application will have one thread with Zeroconf, and a GUI thread that needs to be informed about updates?

This is also possible, but might lead to some duplication.

Tell me if you wish me to go ahead with a proposal for splitting the target class.

stephanemagnenat:
> Tell me if you wish me to go ahead with a proposal for splitting the target class.

I've implemented this proposed change in branch `zeroconf-targetinformation`

stephanemagnenat:
I've merged the `zeroconf-targetinformation` branch (commit a376fa6b923210687e4cdb28c9906b71b7d94236), could you please update the code to match the split, and then I'll merge?

stephanemagnenat:
> I've merged the zeroconf-targetinformation branch (commit a376fa6), could you please update the code to match the split, and then I'll merge?

@davidjsherman could you update the code? I would like to merge this branch as soon as possible as Francesco will have an intern starting from January who would like to test zeroconf. Thank you!

stephanemagnenat:
> I've merged the zeroconf-targetinformation branch (commit a376fa6), could you please update the code to match the split, and then I'll merge?

@davidjsherman could you update the code? I would like to merge this branch as soon as possible as Francesco will have an intern starting from January who would like to test zeroconf. Thank you!

davidjsherman:
I've merged your changes. I will test on Windows and Debian, and rebase.

stephanemagnenat:
What is the status? Were the tests successful? Can I merge?

davidjsherman:
I started a branch [qt-zero](https://github.com/davidjsherman/aseba/tree/qt-zero) for work in progress to use the new classes, but had to stop before I finished. My notes are incomplete but it looks like I am having trouble detecting when all of targets have been completely listed. (Fair warning, the code in the branch is wrong and will generate an EXC_BAD_ACCESS due to unsafe use of std::set.)

davidjsherman:
Here is the diff [f6d6acf..7db9355](https://github.com/davidjsherman/aseba/commit/7db93555fb98d904d8956190b736dfaadaf4e1a4). Target pointers are replaced by complete TargetInformation, which should be right. But then there is some suspicious business about proxying the `xxxCompleted` hooks that seems ill-advised.

stephanemagnenat:
Would you like:
- me to fix to test and try to fix the bug
- wait
?

stephanemagnenat:
I managed to make it work with some changes, see aseba-community/aseba#611.

However, I merged qt-zero with the latest master branch, and there is a lot of additional content pulled it, because branch [zeroconf-PR-derived](/davidjsherman/aseba/tree/zeroconf-PR-derived)  and [qt-zero](/davidjsherman/aseba/tree/qt-zero) also add the scratch client example, latest http changes, and the e2e tests for http.

stephanemagnenat:
I've merged manually recent work on zeroconf, see commit 968c9ff65613f4dbf5f280d43500511375d9f4f9.

stephanemagnenat:
@davidjsherman now that main Zeroconf work has been merged to master, should we close this PR? Or do you prefer to keep it open to discuss further issues, such as tests?

stephanemagnenat:
I think that this PR can be closed now, @davidjsherman ok for you?

PR #559, opened 2016-09-28
Repair OAS apidocs and create handler stubs

davidjsherman:
C++ handler stubs may be created from the OAS apidocs using
```
./apidoc-repair.perl aseba-v1.json | ./apidoc-to-stubs.perl > stubs.cpp
```

The first program removes all examples and moves response descriptions temporarily attached to the schema. The second program maps the corrected endpoints to asebaswitch HTTP handlers. The JSON schema are included as raw strings, with a custom delimiter `#json#` to avoid conflicts with strings ending in ")".

Since Stoplight “helpfully” substitutes `$ref` definitions in the OAS documentation published at https://aseba.api-docs.io/v1, the current API definition is included here as `aseba-v1.json`.


stephanemagnenat:
Makes sense. We should add a small README.md that briefly describes what the perl scripts do.

PR #561, opened 2016-10-05
Playground: Use new Enki viewer with picking

davidjsherman:
This PR extracts from Titwin/aseba-INRIA@e2a61a832ae37955c5715472fa494712c594009b a minimal set of changes to allow Aseba Playground to use the new Enki viewer with picking. The new Enki viewer is not a drop-in replacement for Enki 2.0.0 and changes to client code are needed.

Changes in this PR:
- QTimerEvents must hand off to the new `Enki::ViewerWidget::timerEvent`, which counts milliseconds. If timer events are not handled correctly, picking will not work. A `StartSimulation` hook is used to start the timer.
- `Aseba::PlaygroundViewer` is a QMainWindow with a layout. It is not a derived class of `Enki::ViewerWidget` but does contain one.
- Playground objects with finite mass are flagged as movable by picking.

Note that [Titwin/aseba-INRIA master](https://github.com/Titwin/aseba-INRIA/tree/master) contains other improvements to the GUI that are not included in this PR.

TODO: add a check for the Enki version (which version?) in [targets/playground/CMakeLists.txt](https://github.com/davidjsherman/aseba/blob/playground-picking/targets/playground/CMakeLists.txt). If we want backwards compatibility with Enki 2.0.0, we will need to add conditional compilation in 7 places.

### DIFF
--- a/targets/challenge/challenge.cpp
+++ b/targets/challenge/challenge.cpp
@@ -641,7 +641,6 @@ namespace Enki
 				camera.yaw += 2*M_PI;
 			camera.pitch = -M_PI/7;
 		}
-		ViewerWidget::timerEvent(event);
 	}
 /*
 	void ChallengeViewer::mouseMoveEvent ( QMouseEvent * event )
--- a/targets/playground/PlaygroundViewer.cpp
+++ b/targets/playground/PlaygroundViewer.cpp
@@ -29,6 +29,7 @@
 #include "PlaygroundViewer.h"
 #include "Parameters.h"
 #include "EPuck.h"
+#include "Thymio2.h"
 #include "../../common/utils/utils.h"
 
 #ifdef Q_OS_WIN32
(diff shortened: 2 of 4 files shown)


stephanemagnenat:
I do not think mixing calls to GL outside the Viewer widget is the correct things to do. If I look at this diff I see more code then before, for no clear added benefits. I do not object of using the Viewer inside another Qt application in general, but I think when a simple sub-classing is enough it should be done, and that code specially linked to the viewer such as GL code should be in a subclass.

davidjsherman:
I am uncomfortable with the loss of compatibility.

Most of the lines changed here do not involve picking but the change to use a QMainWindow.
I think that the changes that are strictly necessary for picking are the two lines calling `startTimer` and `timerEvent`. It would be better if this could be moved to Enki, it doesn't feel right to lose backward compatibility for something minor that doesn't really concern the playground.

There was a slightly annoying detail concerning the ‘pickable’ flag, which is now fixed by enki-community/enki@d3f226f89fb9652a3d6699efa9c58464a3e7f9c0. (Previously the flag had to be set by the client code.)

stephanemagnenat:
I agree, I have the same discomfort. With the latest commit on branch `Titwin-picking-squashed` (enki-community/enki@d3f226f89fb9652a3d6699efa9c58464a3e7f9c0), compatibility should be restored.

stephanemagnenat:
This PR is not necessary any more as enki-community/enki@02fdf5b4a8d3bfec8af80bb86cbace0a553c30b2 adds picking while maintaining the API compatibility.

davidjsherman:
I will test compiling Aseba on all OSs with enki-community/enki@02fdf5b and close this PR if all goes well

stephanemagnenat:
This PR is merged.

PR #567, opened 2016-11-02
PR-aseba-enki

Titwin:
Change to support:
- 3d model,
- led color changes,
- clicked interaction,

### DIFF
--- a/targets/playground/Thymio2-natives.cpp
+++ b/targets/playground/Thymio2-natives.cpp
@@ -105,15 +105,6 @@ extern "C" void PlaygroundThymio2Native_leds_circle(AsebaVMState *vm)
 	const sint16 l6(clampValueTo32(vm->variables[AsebaNativePopArg(vm)]));
 	const sint16 l7(clampValueTo32(vm->variables[AsebaNativePopArg(vm)]));
 
-	ASEBA_UNUSED(l0);
-	ASEBA_UNUSED(l1);
-	ASEBA_UNUSED(l2);
-	ASEBA_UNUSED(l3);
-	ASEBA_UNUSED(l4);
-	ASEBA_UNUSED(l5);
-	ASEBA_UNUSED(l6);
-	ASEBA_UNUSED(l7);
-
 	PlaygroundViewer* playgroundViewer(PlaygroundViewer::getInstance());
 	World* world(playgroundViewer->getWorld());
 	for (World::ObjectsIterator objectIt = world->objects.begin(); objectIt != world->objects.end(); ++objectIt)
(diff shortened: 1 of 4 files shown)


Titwin:
there is a bug in enki
file: thymio2.cpp
line 132.
the corrected one should be :

intensity = std::max(0., std::min(1., intensity));

stephanemagnenat:
I fixed the bug in commit ebfe2af786baf2124797059764150780ef0cf292

davidjsherman:
The LED color changes seem excruciatingly slow. When I run [explorateurV7.aesl](https://aseba.wdfiles.com/local--files/en:thymiobehaviourexplorer/explorateurV7.aesl) in playground [thymio-maze-6x6.playground](https://github.com/aseba-community/aseba/blob/master/targets/playground/examples/thymio-maze-6x6.playground), the simulation is a least six times slower.

stephanemagnenat:
I'll look into that. If necessary, we could imagine to have a dynamic cache of pre-rendered textures.

davidjsherman:
Sometimes, for certain playgrounds, the ground texture is hidden by a white rectangle that also masks the grey bar across the bottom. I have observed this consistently with the Yeti and Otter playgrounds, but never with the Maze 6x6 playground. I am using my own compilation on macOS, [inirobot 0.7.4-alpha](https://github.com/davidjsherman/inirobot-scratch-thymioII/releases/tag/0.7.4-alpha). If anyone else observes this I will open an issue.

targets/playground/Thymio2-natives.cpp
```diff
@@ -131,25 +144,25 @@ static double _fmax(double x, double y)
...
-	const double r(paramToColor(clampValueTo32(vm->variables[AsebaNativePopArg(vm)])));
-	const double g(paramToColor(clampValueTo32(vm->variables[AsebaNativePopArg(vm)])));
-	const double b(paramToColor(clampValueTo32(vm->variables[AsebaNativePopArg(vm)])));
-	const double a(_fmax(_fmax(r, g), b));
-	
+	const sint16 r(clampValueTo32(vm->variables[AsebaNativePopArg(vm)]));
+	const sint16 g(clampValueTo32(vm->variables[AsebaNativePopArg(vm)]));
+	const sint16 b(clampValueTo32(vm->variables[AsebaNativePopArg(vm)]));
+	const sint16 a(_imax(_imax(r, g), b));
+	const double param(1./_imax(_imax(r, g),_imax(1,b)));
```
stephanemagnenat: I do not understand why this computation with a is still necessary. I did it before because I was mixing the colour over the white surface, but now with the texture, cannot we simply call setLedColor() with the r, g, b parameters?

targets/playground/Thymio2-natives.cpp
```diff
@@ -104,8 +104,7 @@ extern "C" void PlaygroundThymio2Native_leds_circle(AsebaVMState *vm)
 	const sint16 l5(clampValueTo32(vm->variables[AsebaNativePopArg(vm)]));
 	const sint16 l6(clampValueTo32(vm->variables[AsebaNativePopArg(vm)]));
 	const sint16 l7(clampValueTo32(vm->variables[AsebaNativePopArg(vm)]));
-	
-	// do nothing for now
+
 	ASEBA_UNUSED(l0);
```
stephanemagnenat: the `ASEBA_UNUSED` macros should be removed.

targets/playground/Thymio2-natives.cpp
```diff
@@ -114,14 +113,28 @@ extern "C" void PlaygroundThymio2Native_leds_circle(AsebaVMState *vm)
...
+			thymio2->setLedIntensity(Thymio2::RING_5, l5/32.);
+			thymio2->setLedIntensity(Thymio2::RING_6, l6/32.);
+			thymio2->setLedIntensity(Thymio2::RING_7, l7/32.);
+			return;
+		}
+	}
 }
 
-static double _fmax(double x, double y)
+static sint16 _imax(sint16 x, sint16 y)
```
stephanemagnenat: My fault for duplicating the max function in the first place, but now we should rather use std::max<sint16, sint16>() instead.

targets/playground/Thymio2-natives.cpp
```diff
@@ -159,23 +172,49 @@ extern "C" void PlaygroundThymio2Native_leds_bottom_right(AsebaVMState *vm)
 	const sint16 r(clampValueTo32(vm->variables[AsebaNativePopArg(vm)]));
 	const sint16 g(clampValueTo32(vm->variables[AsebaNativePopArg(vm)]));
 	const sint16 b(clampValueTo32(vm->variables[AsebaNativePopArg(vm)]));
-	
-	// do nothing for now
+	const sint16 a(_imax(_imax(r, g), b));
+	const double param(1./_imax(_imax(r, g),_imax(1,b)));
+
 	ASEBA_UNUSED(r);
```
stephanemagnenat: Again, variables are now used. This also applies later.

Titwin:
About the "gost" texture on the ground Theo observe the same probleme when he is picking a thymio robot. You can ask him for more info about his setup

Titwin:
It's still needed because you can specify a led color near to black (0,0,0,1)
In this case the led pattern will be black instead of no colored at all
The led alpha symbolise here the led intensity, it' why I need to compute it due to led color.

Titwin:
Ok for me,
(sorry for the delay)

davidjsherman:
Added  #571 *Aseba Playground: ground textures sometimes disappear*

stephanemagnenat:
This PR is merged.

Issue #583, opened 2016-12-01
Native function for fast Aseba array Deque


davidjsherman:
Sequential programming with the Aseba language is fastidious. Because the VM doesn't support blocking operations, preemptive scheduling, or a call stack, the approach in [`thymio_motion.aesl `](https://github.com/davidjsherman/inirobot-scratch-thymioII/blob/master/thymio_motion.aesl) is to use a queue that is polled at 100 Hz by a `motor` event. The queue is implemented using arrays and index variables, with insertion handled by scanning forward through the arrays. This *ad hoc* solution is complicated, fragile, and inefficient.

Native functions to implement a fast vector deque would make it easier to implement sequential operations in Aseba programs, and reduce bytecode size. The implementation could be based on [ArrayDeque: Fast Deque Operations Using an Array](http://opendatastructures.org/ods-java/2_4_ArrayDeque_Fast_Deque_O.html). Four operations would be defined:

* `deque.get(A, B, i)` Copy *len*`B` elements from position `i * `*len*`B` in `A`
* `deque.set(A, B, i)` Assign *len*`B` elements to position `i * `*len*`B` in `A`
* `deque.add(A, B, i)` Insert *len*`B` elements at position `i * `*len*`B` in `A`, shifting the existing contents to make room (Figure 2.3). Elements are lost from the right if `A` is full.
* `deque.remove(A, i, n)` Remove `n` elements at position `i * n` in `A`, shifting the existing contents (Figure 2.3).

(The first two elements of `A` are reserved for the `j` and `n` pointers, which is why `get` and `set` aren't simply implemented using existing Aseba vector addressing operations.)

Implementing a simple motion queue would give something like:
```
var Queue[122] # Store up to 40 operations e.g. for BeeBot
var operation[3]

sub motion_add
call deque.add(Queue,operation[0:2],-1) # push at back

onevent timer0
call deque.get(Queue,operation,0) 
timer.period[0] = operation[0]
motor.left.target = operation[1]
motor.right.target = operation[2]
deque.remove(Queue,0,3) # pop from front
```

stephanemagnenat:
I like it! @mbonani do we still have some flash space left in the Thymio firmware to add some native functions?

stephanemagnenat:
About the API, I understand that negative positions are counting from the back, which is not supported in the referenced document, right?

Do you also plan to have the operations `push_back`, `push_front`, `pop_back`, and `pop_front`, for ease of use?

[...]

PR #585, opened 2016-12-04
Native function for fast array Deque

davidjsherman:
This PR defines native functions for fast deque operations on Aseba arrays, as discussed in #583. It contains two parts:
1. Unit tests for the deque operations, defined in a behavior drive development style. The tests are performed by checking the VM memory after each step, using `asebatest` as the test runner.
1. An implementation of the basic operations `deque.get`, `deque.set`, `deque.insert`, and `deque.erase`.

This initial implementation is correct (according to the tests), but not optimized, neither for speed nor for code size. It does not include the convenience functions `deque.push_front`, `deque.push_back`, `deque.pop_front` nor `deque.pop_back`.

The tests are defined in the `tests/vm` directory introduced by [PR 584](https://github.com/aseba-community/aseba/pull/584) and this PR is meant to be merged after that one.


stephanemagnenat:
Why is this PR still reflecting to master before the unit tests refactoring was merged? Does github requires a manual operation?

davidjsherman:
I just force-pushed a rebase, it looks all right now.

tests/vm/CMakeLists.txt
```diff
@@ -4,3 +4,19 @@ add_executable(aseba-test-natives-count
 )
 target_link_libraries(aseba-test-natives-count asebacompiler asebavm asebavmdummycallbacks ${ASEBA_CORE_LIBRARIES})
 add_test(natives-count ${EXECUTABLE_OUTPUT_PATH}/aseba-test-natives-count)
+
+# test the deque native functions
+add_test(NAME deque-empty COMMAND $<TARGET_FILE:asebatest> --memcmp
```
davidjsherman: No, not necessary, CMake can find the target `asebatest`. However there is a problem with the `simulate.py` tests in `tests/compiler`, CMake isn't generating the correct value of `${CMAKE_CURRENT_BINARY_DIR}` for Xcode. I was trying a global fix a generator time but it didn't work

vm/natives.c
```diff
@@ -1138,5 +1138,393 @@ const AsebaNativeFunctionDescription AsebaNativeDescription_rand =
...
+	uint16 src = AsebaNativePopArg(vm);
+	uint16 index = AsebaNativePopArg(vm);
+	uint16 index_val = vm->variables[index];
+	
+	// variable size
+	uint16 dest_length = AsebaNativePopArg(vm);
+	uint16 src_length = AsebaNativePopArg(vm);
+	
+	// Check for exception
+	if (vm->variables[index] < 0 || index_val > dest_length - 2 - 1
```
stephanemagnenat: Same comment as for line 1231

vm/natives.c
```diff
@@ -1138,5 +1138,393 @@ const AsebaNativeFunctionDescription AsebaNativeDescription_rand =
...
+	uint16 index = AsebaNativePopArg(vm);
+	uint16 len = AsebaNativePopArg(vm);
+	uint16 index_val = vm->variables[index];
+	uint16 len_val = vm->variables[len];
+	
+	// variable size
+	uint16 dest_length = AsebaNativePopArg(vm);
+	
+	// Check for exception
+	if (vm->variables[index] < 0 || index_val > dest_length - 2 - 1
```
stephanemagnenat: Same comment as for line 1231

vm/natives.c
```diff
@@ -1138,5 +1138,393 @@ const AsebaNativeFunctionDescription AsebaNativeDescription_rand =
...
+	uint16 index = AsebaNativePopArg(vm);
+	uint16 index_val = vm->variables[index];
+	
+	// variable size
+	uint16 dest_length = AsebaNativePopArg(vm);
+	uint16 src_length = AsebaNativePopArg(vm);
+	
+	// Check for exception
+	if (vm->variables[index] < 0 || index_val > dest_length - 2 - 1
+		|| src_length > dest_length - 2 - vm->variables[dest])
```
stephanemagnenat: After thinking for a while, I understand that `vm->variables[dest]` is the index. I think that it would make the code more readable to put it explicitly in a temporary variable. But maybe a comment is sufficient.

vm/natives.c
```diff
@@ -1138,5 +1138,393 @@ const AsebaNativeFunctionDescription AsebaNativeDescription_rand =
...
+void AsebaNative_deqpushfront(AsebaVMState *vm)
+{
+	// Pop arguments off
+	uint16 dest = AsebaNativePopArg(vm);
+	uint16 src = AsebaNativePopArg(vm);
+	uint16 dest_length = AsebaNativePopArg(vm);
+	uint16 src_length = AsebaNativePopArg(vm);
+
+	// Now call insert with proper arguments
+	_AsebaNative_deqinsert(vm, dest, src, 0, dest_length, src_length);
```
stephanemagnenat: So there is no check for empty space for push_front/back operations? I think that it would be nice, because they are the high-level API.

vm/natives.c
```diff
@@ -1138,5 +1138,393 @@ const AsebaNativeFunctionDescription AsebaNativeDescription_rand =
...
+void AsebaNative_deqpopfront(AsebaVMState *vm)
+{
+	// Pop arguments off
+	uint16 dest = AsebaNativePopArg(vm);
+	uint16 src = AsebaNativePopArg(vm);
+	uint16 dest_length = AsebaNativePopArg(vm);
+	uint16 src_length = AsebaNativePopArg(vm);
+
+	// call insert and erase with proper arguments
+	_AsebaNative_deqget(vm, dest, src, 0, dest_length, src_length);
```
stephanemagnenat: I would add a check for underflow for pop_front/back, as these are the high-level API.

vm/natives.c
```diff
@@ -1138,5 +1138,393 @@ const AsebaNativeFunctionDescription AsebaNativeDescription_rand =
...
+	uint16 src = AsebaNativePopArg(vm);
+	uint16 index = AsebaNativePopArg(vm);
+	uint16 index_val = vm->variables[index];
+	
+	// variable size
+	uint16 dest_length = AsebaNativePopArg(vm);
+	uint16 src_length = AsebaNativePopArg(vm);
+	
+	// Check for exception
+	if (vm->variables[index] < 0 || index_val > src_length - 2 - 1
```
davidjsherman: Because I want to throw an exception if the supplied argument is < 0, while declaring `index_val` unsigned as it should be. Since an exception is raised in `index_val` is larger than the array, the silent limitation of `index_val` to [0,32767) isn't a constraint.

vm/natives.c
```diff
@@ -1138,5 +1138,393 @@ const AsebaNativeFunctionDescription AsebaNativeDescription_rand =
...
+		|| dest_length > vm->variables[src] - index_val)
+	{
+		vm->flags = ASEBA_VM_STEP_BY_STEP_MASK;
```
davidjsherman: Yes, that would save some bytes from the code size.

vm/natives.c
davidjsherman: The check is performed in `insert`, it uses the tuple size to check that there is enough space. It is checked in `tests/vm/data/deque-err-insert-toobig.txt`.

vm/natives.c
```diff
@@ -1138,5 +1138,393 @@ const AsebaNativeFunctionDescription AsebaNativeDescription_rand =
...
+static void deque_shift(AsebaVMState *vm, uint16 dq, uint16 dq_capacity, uint16 target, sint16 last, sint16 delta)
+{
+	for ( ; (delta < 0 ? target >= last : target <= last) ; (delta < 0 ? target-- : target++) )
+		vm->variables[dq + 2 + (target % dq_capacity)] = vm->variables[dq + 2 + ((target + delta) % dq_capacity)];
+}
+
+void AsebaNative_deqsize(AsebaVMState *vm)
+{
+	// variable pos
+	uint16 size = AsebaNativePopArg(vm);
```
stephanemagnenat: No they always put it first, such as dest <- op(args...).

But for me the deque is different, I fill it is a pseudo-object construct, so I would always put the this first.

vm/natives.c
davidjsherman: Unless we introduce exceptions especially for the high-level API, these checks are already performed by the low-level functions. We could explicitly add error tests for the high-level functions to make sure that nothing is lost in translation.

vm/natives.c
stephanemagnenat: Oh this makes sense! But I did not get it before, maybe you can add a comment? ;-)

vm/natives.c
davidjsherman: Sure, all right

vm/natives.c
davidjsherman: Actually you were right, the capacity check was only in the façade! Fixed in af131f9fc220c34ae98a498ac978bf3429a38313, which also has error tests for `push` and `pop`.

vm/natives.c
davidjsherman: Thinking about it, I agree that systematically putting deque first is familiar for us, but is it true for the young programmers in Aseba's target audience? If the rest of the native functions consistently use the first argument as the output, should we break that consistency because *we* think of the deque as a object?

Unless there is a pedagogical goal of introducing object-oriented thinking, and in the future there might be other object-oriented native functions ?

davidjsherman:
All current review comments, except for the [parameter order question](https://github.com/aseba-community/aseba/pull/585#discussion_r91940075), addressed in 1ecde4c7608a23ea4eaa5187119c6b41af9a1642.
Also improved tests and some bug fixes.

vm/natives.c
stephanemagnenat: It is a very good point. Until now the prefixes in Aseba stdlib are conceptually modules, such as "math.*". In the case of deque, we can continue to see it as a module, in case we should have destination first, or see it as a pseudo-object, in case we should have the _this_ pointer first.

It is true that personally I like the pseudo object, and that on the one hand it could be interesting to teach this concept ; but in the other hand, maybe it is adding a new convention for a special case, which is bad. Maybe the best is to write down the different options, look at them, and take the one that _feels_ the right one?

davidjsherman:
Based on our discussion, I reordered the arguments to use an object-oriented style.

Commit c3e3ee66692deee368fba271896bbac2ac38d8a0 reorders the arguments in the tests, then fa420fcde51d1b7fc4bdab8c1dc4f0a11d2d4d08 reorders the arguments in the implementation. Revision 2 of the [preliminary documentation page](https://www.thymio.org/en:asebastdnativedeque) reflects this reordering.

stephanemagnenat:
This PR looks good to merge. Do you want to rebase it so that it appears as a single comment instead of keeping the whole history of our experiments?

Regarding the documentation page, do you think it could be integrated in the main one once the firmware with deque support is released, or do you think they should be kept separate? Currently the offline documentation is built by parsing the Wiki. If we use a sub-page we will have to modify the script. We can of course do it, if we think it is useful.

davidjsherman:
This branch is now rebased into one commit f0d5bb5a487976ace02a3c2a3dc606838fe10710.

I think that the nine deque functions should be added to the main documentation [Native functions standard library](https://www.thymio.org/en:asebastdnative), but that the detailed documentation in [Native functions deque library](https://www.thymio.org/en:asebastdnativedeque) with examples could be kept in the Wiki.

Some ideas for examples: binary search, limited-depth fractals, improved line following with a simple Markov chain.

stephanemagnenat:
> This branch is now rebased into one commit f0d5bb5.

Thank you! merged :-)

> I think that the nine deque functions should be added to the main documentation Native functions standard library, but that the detailed documentation in Native functions deque library with examples could be kept in the Wiki.

I agree, I am also considering structuring the native functions library into sections, for instance:
- math: vector operations
  most of current functions  
- math: algorithm
  `math.sort` and `math.rand` (prefixing them math was a bad choice, I am still thinking whether I should rename them `algo.`)
- container: deque
  `deque.*`

> Some ideas for examples: binary search, limited-depth fractals, improved line following with a simple Markov chain.

I like a lot these ideas, do you see them in the detailed documentation?

davidjsherman:
I am closing #583

PR #613, opened 2017-03-31
Readme: aseba.wikidot.com -> www.thymio.org

Grahack:


stephanemagnenat:
I am using  the non-Thymio name on purpose, as this is a README for Aseba in general, not for Thymio in particular.

PR #640, opened 2017-05-30
Rename Enki polygone to polygon

davidjsherman:
Depends on enki-community/enki#45

### DIFF
--- a/targets/playground/Door.cpp
+++ b/targets/playground/Door.cpp
@@ -101,7 +101,7 @@ namespace Enki
 	
 	// AreaActivating
 	
-	AreaActivating::AreaActivating(Robot *owner, const Polygone& activeArea) :
+	AreaActivating::AreaActivating(Robot *owner, const Polygon& activeArea) :
 		activeArea(activeArea),
 		active(false)
 	{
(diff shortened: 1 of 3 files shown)


mbonani:
Hello
It seems it brakes compilation on 1.5.x branch:

stephanemagnenat:
> It is perhaps not important but just want to warn it there is something to do...

Thanks for reporting, it is relevant because it does break the compilation instructions. The underlying reason is that Aseba depends on a development branch of Enki that thus does not have a stable API. I know where I want to go with Enki (ECS architecture, see https://github.com/enki-community/enki/tree/ecs-refactor), but lack engineering resources to progress at a good pace, hence the dependency on the unstable API.

There are three ways to fix that:
1. Update Aseba in the 1.5.x branch to match latest Enki API. _Drawback:_ Requires a bit of back-porting work.
2. Always use master instead of latest stable branch in compilation instructions. _Drawback:_ Potentially let people compile totally broken versions of Aseba.
3. Use tags for Enki in compilation instructions, to ensure fetching a version that has a compatible API. _Drawback:_ Must update the tags on Enki side every time we change something. A solution could be to have one branch on Enki's side for every API used by Aseba.
4. Use git submodules to track specific versions of Enki and Dashel in Aseba. This is nice for standalone compilation, but not correct from a system-wide library installation perspective. Indeed, on Linux one wish that Dashel is compiled as a dynamic library, so that it can be updated, for instance for security, without having to update Aseba.

All things considered, I think that 1 is the easiest solution in short term.

stephanemagnenat:
I think that now as 1.6 release is almost here, this is not a problem any more.

PR #653, opened 2017-06-29
WIP: Zeroconf unit testing

davidjsherman:
For discussion. Unit tests in [`test-zeroconf.cpp`](https://github.com/aseba-community/aseba/compare/master...davidjsherman:zeroconf-unit-tests?expand=1#diff-da707edeb9e6764b1e401d016aea5946) defined with Catch.
This PR will not compile in master but will if merged into e.g. 6d415e162ae8c35aef6036dfd8a87ce42ef99fa0. 

- [ ] unit testing of the Zeroconf container
- [ ] advertise and browse local targets ⚠️ 
- [ ] browse remote targets
- [ ] stop advertising a local target

:warning: Since the basic Aseba::Zeroconf object is single-threaded, it can't advertise and browse at the same time!

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -2,6 +2,7 @@ add_subdirectory(msg)
 add_subdirectory(compiler)
 add_subdirectory(vm)
 add_subdirectory(simulator)
+add_subdirectory(zeroconf)
 
 
 # test asebahttp
--- a/tests/catch.hpp
+++ b/tests/catch.hpp
@@ -1,0 +1,10531 @@
+/*
+ *  Catch v1.5.9
+ *  Generated: 2016-11-29 12:14:38.049276
+ *  ----------------------------------------------------------
+ *  This file has been merged from multiple headers. Please don't edit it directly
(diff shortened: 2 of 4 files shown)


tests/CMakeLists.txt
```diff
@@ -2,6 +2,7 @@ add_subdirectory(msg)
 add_subdirectory(compiler)
 add_subdirectory(vm)
 add_subdirectory(simulator)
+add_subdirectory(zeroconf)
```
stephanemagnenat: This should be guarded with `if (DNSSD_FOUND)`.

stephanemagnenat:
This PR is not against master right? Because in master the Zeroconf object contains targets while as I understand the test assumes Zeroconf is a list of targets.

davidjsherman:
Yes, these tests are defined using the examples, so against the API in zeroconf-PR-derived. I would like to extract the interesting parts of the examples and make tests from them, to help choose the right API.

I will add testing programs for each of the derived classes.
Unfortunately it looks like the basic implementation (without threads, dashel or Qt) can't be usefully tested because it is single-threaded.

PR #680, opened 2017-08-11
Asebaswitch will advertise itself using Zeroconf

davidjsherman:
If Zeroconf is configured, advertise the switch with PID `ASEBA_PID_UNDEFINED`

The constructor `Switch::Switch` now takes a name, by default "Aseba Switch".

TODO: If the switch sees node descriptions go past, it could update its TXT record to reflect the PIDs it has available.

### DIFF
--- a/switches/switch/CMakeLists.txt
+++ b/switches/switch/CMakeLists.txt
@@ -2,7 +2,7 @@ add_executable(asebaswitch
 	switch.cpp
 )
 
-target_link_libraries(asebaswitch ${ASEBA_CORE_LIBRARIES})
+target_link_libraries(asebaswitch ${ASEBA_ZEROCONF_LIBRARIES} ${ASEBA_CORE_LIBRARIES})
 
 install(TARGETS asebaswitch RUNTIME
 	DESTINATION bin
--- a/switches/switch/switch.cpp
+++ b/switches/switch/switch.cpp
@@ -32,6 +32,10 @@
 #include "../../common/utils/utils.h"
 #include "../../common/msg/msg.h"
 #include "../../common/msg/endian.h"
+#ifdef ZEROCONF_SUPPORT
+#include "../../common/zeroconf/zeroconf-dashelhub.h"
+#include "../../common/productids.h"
+#endif // ZEROCONF_SUPPORT
 
 namespace Aseba 
 {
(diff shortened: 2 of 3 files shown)


switches/switch/CMakeLists.txt
```diff
@@ -2,7 +2,7 @@ add_executable(asebaswitch
 	switch.cpp
 )
 
-target_link_libraries(asebaswitch ${ASEBA_CORE_LIBRARIES})
+target_link_libraries(asebaswitch ${ASEBA_ZEROCONF_LIBRARIES} ${ASEBA_CORE_LIBRARIES})
```
stephanemagnenat: The variable `ASEBA_ZEROCONF_LIBRARIES` is not defined if `DNSSD_FOUND` is false, so I think this will produce an error. In Studio, this is solved by having an `ADDITIONAL_LIBRARIES` variable that accumulates all optional library dependencies. As zeroconf will be used in different places, this is maybe too verbose. A way to do it cleanly is to define a variable called `ASEBA_OPTIONAL_ZEROCONF_LIBRARIES`, whose value is either `ASEBA_ZEROCONF_LIBRARIES` if `DNSSD_FOUND` is true, or empty otherwise.

switches/switch/switch.cpp
```diff
@@ -42,18 +46,34 @@ namespace Aseba
 	/*@{*/
 
 	//! Broadcast messages form any data stream to all others data streams including itself.
-	Switch::Switch(unsigned port, bool verbose, bool dump, bool forward, bool rawTime) :
+	Switch::Switch(unsigned port, std::string name, bool verbose, bool dump, bool forward, bool rawTime) :
```
stephanemagnenat: Please pass name as const reference.

switches/switch/switch.h
```diff
@@ -43,8 +46,8 @@ namespace Aseba
 				@param dump should we dump content of each message
 				@param forward should we only forward messages instead of transmit them back to the sender
 			*/
-			Switch(unsigned port, bool verbose, bool dump, bool forward, bool rawTime);
-			
+			Switch(unsigned port, std::string name, bool verbose, bool dump, bool forward, bool rawTime);
```
stephanemagnenat: Please pass name as const reference.

stephanemagnenat:
> TODO: If the switch sees node descriptions go past, it could update its TXT record to reflect the PIDs it has available.

This should be done in the new http2-based switch, not this one, as this one does not embed a `NodesManager`.

stephanemagnenat:
I'm now doing the merge by hand, integrating code from http as well.

stephanemagnenat:
Merged through f0b2c136a252a12448969e9797be4370b700e9a9.

Issue #674, opened 2017-08-07
Native sort function has too high complexity


stephanemagnenat:
The native function `math.sort` is using a [Comb sort algorithm](https://en.wikipedia.org/wiki/Comb_sort) which has a O(n^2) worst-case performance. This should be improved with a better sort algorithm.

PR #688, opened 2017-09-18
Added heap sort, tested against aseba_comb_sort and C++ std::sort

shilingwang:
As described in issue #674, the complexity of `math.sort` (comb sort) is worst case O(n^2) and best case O(n log(n)). Here I implement heap sort, which should be worst-case O(n log(n)) complexity. The test results (i7-5600U CPU) for comb sort, heap sort and c++ std::sort show that when the data are in random order, the three implementations are not much different and std::sort and comb sort are always a little bit better than heap sort. When the data are already in sorted order without duplication, heap sort is slower than the other two (no more than double). However, when the data is const, heap sort is much faster than comb sort (10 times or more).

### DIFF
--- a/tests/test_sort.cpp
+++ b/tests/test_sort.cpp
@@ -1,0 +1,210 @@
+#include <iostream>
+#include <vector>
+#include <map>
+#include <functional>
+#include <chrono>
+#include <ctime>
+#include <algorithm>
+
+using namespace std;
+
+// sorting functions
+
+// comb sort ( from Wikipedia )
+void aseba_comb_sort(int16_t* input, uint16_t size)
+{
+	uint16_t gap = size;
+	uint16_t swapped = 0;
+	uint16_t i;
+
+	while ((gap > 1) || swapped)
+	{
+		if (gap > 1)
+		{
+#ifdef __C30__
+			gap = __builtin_divud(__builtin_muluu(gap,4),5);
+#else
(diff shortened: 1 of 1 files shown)


stephanemagnenat:
Here (macOS 10.12.6, clang 8.1.0, Core i7-6700K, -O2), `std::sort` outperforms everything else in any condition. For 1k element arrays, heap sort is 4 times faster than comb sort on constant arrays, 4 times slower on already-sorted arrays, and about the same speed otherwise.

It would be very interesting to test on the typical target processor of Aseba, I expect heap sort to be proportionally faster when the processor becomes more primitive.

stephanemagnenat:
Do we have a way to run performance test on a simulator of the target's device?

davidjsherman:
The Microchip simulator has a Stopwatch that can measure the number of milliseconds between breakpoints. So it should be possible to test the timing in the MPLABX IDE simulator.

While it is possible to submit a script of commands to Microchip's command-line simulator `sim30`, as far as I can tell the stopwatch isn't available.

If you write an Aseba language program that tests the sorting speed, for example by incrementing a counter with every `onevent motor`, then I suppose one could run that bytecode in the Microchip simulator. The `sim30` script could write the contents of the counters to the pseudo-UART, so that it wouldn't be necessary to analyze a VM memory dump.

stephanemagnenat:
That's a good idea. Actually we could also try with the physical robot. One just need to recompile the firmware.

PR #722, opened 2017-11-20
Mask conflicting symbol from mingw32 corerror.h (Windows error codes)

davidjsherman:
Windows error codes in /mingw32/i686-w64-mingw32/include/corerror.h also
define ERROR_STACK_OVERFLOW, which conflicts with symbols defined by
the Aseba compiler. Mask the Windows symbol that I think we don't need.

### DIFF
--- a/compiler/errors_code.h
+++ b/compiler/errors_code.h
@@ -2,6 +2,11 @@
 #ifndef __ERRORS_H
 #define __ERRORS_H
 
+// Mask conflict with /mingw32/i686-w64-mingw32/include/corerror.h (Windows error codes)
+#ifdef WIN32
+#undef ERROR_STACK_OVERFLOW
+#endif
+
 /*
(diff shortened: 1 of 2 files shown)


stephanemagnenat:
It seems we are fixing problems at the same time ;)

Commit 9174feaec49838e0cf0b968325ac84dd81b1e997 uses `WIN32_LEAN_AND_MEAN` define to fix a conflict with the Win32 macro `REGISTERED` which I use in `Zeroconf::Target::State`. Does it also fixes the error this patch fixes?

stephanemagnenat:
@davidjsherman is this PR still relevant (given my comment above)?

davidjsherman:
This change doesn't appear to be necessary on Windows 7, either, according to https://ci.inria.fr/inirobot/job/aseba-community/job/aseba/job/master/28/.

PR #723, opened 2017-11-23
menu targets Dashel improvement for documentation and example

mbonani:


stephanemagnenat:
Thank you for the code, this addition is nice.

However, this PR breaks existing code because the patch is not against the tip of master, but against an older version.

So I am merging by hand.

stephanemagnenat:
Merged by hand in dc16864.

PR #744, opened 2017-12-12
Thymio default behavior improvement

mbonani:


stephanemagnenat:
Merged by hand through cherry-picking in f39f4055.

Issue #705, opened 2017-10-13
Provide an AESL file containing the default behaviour of Thymio


stephanemagnenat:
This is the pendant of #624, but by loading the file by hand.

stephanemagnenat:
@davidjsherman had the idea that if massloader can launch a file once, then the existing mechanism of playground to launch third-party command could be used to load a file by default. It is not ideal as it would show some garbage in the log of playground, but it could be a temporary fix.

stephanemagnenat:
Commit 3d104508 adds support for loading a code only once with asebamassloader.

mbonani:
[all-behavior-for-simulator.zip](https://github.com/aseba-community/aseba/files/1495381/all-behavior-for-simulator.zip)

Here is all behavior that are usable in the simulator. I let the attentive mode only in the menu but it do nothing.  Where should I add it in aseba? Do we add it only in packaging as a dependency?

stephanemagnenat:
I will fix some minor things (indentation, space consistency) and add it the the playground examples. Then we can try to let Playground load it through asebamassloader. We should make sure in the packaging that asebamassloader is available in the same directory as asebaplayground.

stephanemagnenat:
I cleaned the code up and added it, see commit 91c1ab9e.

@mbonani could you please further test? I feel that with the explorer behaviour sometimes the robot turns too fast and stays too far away from obstacle.

mbonani:
The motor target where not bounded, I correct this in my pool request.
But there is still a difference with the C code because in will not overflow like the aesl code:

```
	long temp1 = 0;
		long temp2 = 0;
		temp1 += vmVariables.prox[0];
		temp1 += vmVariables.prox[1] * 2;
		temp1 += vmVariables.prox[2] * 3;
		temp1 += vmVariables.prox[3] * 2;
		temp1 += vmVariables.prox[4];
		temp2 += vmVariables.prox[0] * -4;
		temp2 += vmVariables.prox[1] * -3;
		temp2 += vmVariables.prox[3] * 3;
		temp2 += vmVariables.prox[4] * 4;
		vmVariables.target[0] = speed - __builtin_divsd((temp1 + temp2) * speed, 2000);
		vmVariables.target[1] = speed - __builtin_divsd((temp1 - temp2) * speed, 2000);
		
		if(vmVariables.target[0] < -600) 
			vmVariables.target[0] = -600;
		if(vmVariables.target[1] < -600)
			vmVariables.target[1]= -600;
[...]

PR #754, opened 2017-12-18
fix thymio-default-behaviours.aesl

mbonani:
now it is like the C firmware code


stephanemagnenat:
Thank you! It means we can close #705 now right?

PR #755, opened 2017-12-19
Handle reconnecting a target while the Web Server is running

cor3ntin:
Handle disconnecting and reconnecting a device in `asebahttp`.
Only works for targets associated with a constant node id across reconnection.
Typically it will works great for Thymio, less so for simulated robots.
The intent is to offer a workable solution for the 1.6 release, not to be a long term solution.

### DIFF
--- a/switches/http/http.cpp
+++ b/switches/http/http.cpp
@@ -160,47 +160,23 @@ namespace Aseba
 #endif // ZEROCONF_SUPPORT
     // created empty: pendingResponses, pendingVariables, eventSubscriptions, httpRequests, streamsToShutdown
     {
+        for (strings::const_iterator it = targets.cbegin(); it != targets.end(); ++it)
+        {
+            streamInitParameters.insert(std::pair<std::string, Dashel::Stream*>(*it, NULL));
+        }
         // listen for incoming HTTP and Aseba requests
         if (! inHttpPort.empty())
             inHttpStream = connect("tcpin:port=" + inHttpPort);
         if (! inAsebaPort.empty())
             inAsebaStream = connect("tcpin:port=" + inAsebaPort);
-
-
-        // connect to each Aseba target
-        for (strings::const_iterator it = targets.begin(); it != targets.end(); it++)
-        {
-            try {
-                if (verbose)
-                    std::cout << "HttpInterface connect asebaTarget " << *it << "\n";
-                if (Dashel::Stream *cxn = connect(*it)) // triggers connectionCreated
-                {
-                    asebaStreams[cxn].clear(); // no node id until description is received
-                    Dashel::ParameterSet parser;
-                    parser.add("dummy:remapLocal=0;remapTarget=0;remapAesl=0");
-                    parser.add((*it).c_str());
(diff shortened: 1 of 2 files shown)


switches/http/http.cpp
```diff
@@ -263,35 +239,16 @@ namespace Aseba
...
-                cerr << "Last dashel connection closed, stopping hub" << endl;
-                stop();
-            }
-        }
-        else                                 // this is the HTTP stream
-        {
+            discard_stream(stream);
             if (verbose)
-                cerr << stream << " Connection closed to " << stream->getTargetName() << endl;
+            cout << stream << " Connection closed to " << stream->getTargetName() << endl;
```
stephanemagnenat: missing indentation

switches/http/http.cpp
```diff
@@ -263,35 +239,16 @@ namespace Aseba
...
             unscheduleAllResponses(stream);
             pendingResponses.erase(stream);
             unsigned num = streamsToShutdown.erase(stream);
             if (verbose)
-                cerr << stream << " Connection closed, removed " << num << " pending shutdowns" << endl;
-        }
+            cout << stream << " Connection closed, removed " << num << " pending shutdowns" << endl;
```
stephanemagnenat: missing indentation

switches/http/http.cpp
```diff
@@ -1560,6 +1571,31 @@ namespace Aseba
...
+
+        for (std::map<std::string, Dashel::Stream*>::iterator it = stream_init_parameters.begin();
+            it != stream_init_parameters.end(); ++it)
+        {
+            if (it->second == stream)
+                it->second = NULL;
+        }
+    }
+
+    std::string HttpInterface::targetFromString(Dashel::Stream* stream)
```
stephanemagnenat: Shouldn't this method be const?

switches/http/http.cpp
```diff
@@ -1560,6 +1571,31 @@ namespace Aseba
             return newId;
         }
     }
+
+    void HttpInterface::discard_stream(Dashel::Stream* stream) {
```
stephanemagnenat: Wrong naming convention, use `discardStream`

switches/http/http.cpp
```diff
@@ -1560,6 +1571,31 @@ namespace Aseba
...
+        //closeStream(stream);
```
stephanemagnenat: I would add a comment "called from connectionClosed"

switches/http/http.h
```diff
@@ -72,7 +72,8 @@ namespace Aseba
         typedef std::map<unsigned, std::wstring>                NodeIdProgramMap;
 
     protected:
-        // streams
+        std::map<std::string, Dashel::Stream*> stream_init_parameters;
```
stephanemagnenat: Wrong naming convention, use `streamInitParameters`.

switches/http/http.h
```diff
@@ -72,7 +72,8 @@ namespace Aseba
...
+        std::map<std::string, unsigned> targets_to_nodeid;
```
stephanemagnenat: Wrong naming convention, use `targetsToNodeId`.

PR #761, opened 2017-12-20
Load AESL using persistent asebahttp process

davidjsherman:


mbonani:
If we agree on http we should not change the default http port so that blockly and scrtachx connect 
Also keeping 33333 on Thymio and http make it appears only once on the aseba studio connection window. It test that: 

<process command=":asebahttp --autorestart -s 33333 -a thymio-default-behaviours.aesl tcp:localhost;33333" />

otherwise the correction on asebamassloader is working

cor3ntin:
I vote for not merging that for 1.6, it's a bigger conversation.

mbonani:
I propose to make simple for this prerelease. We keep massloader on most world and we propose only one word with the challenge pack that as a http and it has web bridge in its name. We have so a solution to propose to people for testing.

davidjsherman:
I agree.

Issue #760, opened 2017-12-19
asebamassloader is not stable on windows in playground


mbonani:
I notice instability with asebamassloader on my computer with the simulator. I test with latest build and yeti discovery kits. 
Few times it works message on simulator consol are correct and behavior is charged

![capture-sim0](https://user-images.githubusercontent.com/1000934/34181791-ee4b8bd0-e513-11e7-889a-8eef9ee36ab5.PNG)

Some time message are correct (perhaps order is wrong) and behavior is not launched:

![capture-sim1](https://user-images.githubusercontent.com/1000934/34181850-2ede24be-e514-11e7-91b5-1f5732b67431.PNG)

Most time message are showing problems: 
![capture-sim2](https://user-images.githubusercontent.com/1000934/34181866-3edee7fe-e514-11e7-9dcf-ec3f84d3afe1.PNG)

There is probably a timing issue

davidjsherman:
Does it work if you  replace `asebamassloader` by `asebahttp` in the playground file, and hide the robot behind the `asebahttp` switch? (The switch would run permanently on port 33333, with only one robot node in its network on, say, port 33334.)

Because with `asebahttp` one is guaranteed that the AESL program is fully compiled and loaded, before the switch functionality is started.

stephanemagnenat:
I agree that it is likely a timing issue. @davidjsherman's approach would work but require running http yor the duration of the session (which could be a good think from a blockly/scratch perspective).

Another approach could be to add some delays to the process option to let it launch later.

I guess that for 1.6, the http approach is probably easier.

stephanemagnenat:
Looking at the error in the last screenshot and at the medula code (where there is a one second delay after a successful connection until the code is uploaded), I am not so sure that it is a delay problem related to the description of the target. On the faulty screenshot, the code seems indeed to be loaded properly.

stephanemagnenat:
After some thought, I believe that the problem is that massloader would just close after the Run() message had been sent and the stream flushed, but on Win32, stream::flush() does not ensure a flush. I added another 1s second of wait in 0d2927bf, hopefully it solves the problem.

mbonani:
I will test,
but yes it is working with asebahttp, it is also interesting to have the simulated Thymio ready with web bridge. It could simplify for the user. 

Also should we load each time the default behavior on all Thymio (in the R2T2 world)?
Only on some world? in that case should we have a naming convention at the file to know? Should we wait planned improvement of 1.7?

[...]

PR #762, opened 2017-12-20
Load default behavior for all Thymios

davidjsherman:
Suggested by https://github.com/aseba-community/aseba/issues/760#issuecomment-352996423


davidjsherman:
Compared to #761, the R2T2 worlds are a case where Massloader is a better approach, because we want all of the 4 or 16 Thymios to be individually addressable. Using Http would join them into a single network, controlled by a single client.

mbonani:
thanks

PR #765, opened 2017-12-20
fix thymio challenge  pack world

mbonani:


targets/playground/examples/thymio-challenge-pack-web-bridge.playground
```diff
@@ -2,14 +2,20 @@
...
 	<color name="orange" r="1" g="0.3" b="0" />
 
 	<world w="104.1" h="79.4" color="white" groundTexture="piste-thymio-challenge-pack.svgz" />
 	<wall x="50" y="35" l1="10" l2="5" h="5" mass="10" color="orange" angle="0.000" />
-	<robot type="thymio2" x="70" y="15" port="33333" name="Thymio" />
+	<robot type="thymio2" x="70" y="15" port="33330" name="Thymio" />
 	
-	<process command=":asebahttp -n ThymioChallenge -v  --autorestart -s 33333 tcp:localhost;33333" />
-
+	<process command=":asebamassloader --once thymio-default-behaviours.aesl tcp:localhost;33333" />
```
stephanemagnenat: So mass loader connects through http, it probably works but this is getting quite complicated...

stephanemagnenat: Why do we need asebamassloader as http can load a file by itself (through the `--aesl` or `-a` option) ?

davidjsherman: Use the `--aesl` option of asebahttp to load the AESL file without asebamassloader, nor remapping the Aseba target port:
```
<process command=":asebahttp --autorestart --aesl thymio-default-behaviours.aesl tcp:localhost;33333" />
```

mbonani: I tested and it is not working anymore. Probably last changes on http breaks something.

mbonani:
I tested several configuration and that one was working with actual http. I wanted to discuss with Corentin about that. I didn't get all implication that was discuss in #761 #760.

targets/playground/examples/thymio-challenge-pack-web-bridge.playground
mbonani: i tested and it was not working now, probably last chages on http breaks someting.

davidjsherman:
@mbonani You are being polite we hijacked your thread.
I don't thing that asebahttp can run the Aseba switch on the same port as its target. That is why I didn't specify a port in https://github.com/aseba-community/aseba/pull/765#discussion_r158205583, normally this means that the switch functionality isn't started. Alternatively, you can request a random ephemeral port for the switch using `-s 0`.

targets/playground/examples/thymio-challenge-pack-web-bridge.playground
stephanemagnenat: Ok, then the last change needs to be fixed and the logical solution (asebahttp) be used here.

stephanemagnenat:
If the last commit on http by Corentin breaks the behaviour of asebahttp (to be verified), then it should be fixed. Then we can use asebahttp here. We must not commit hacks just before the release because we introduced bugs. Rather we should delay the release or not include the faulty elements into the release (in that case the release candidate). Otherwise we build technical debt that kills us on the longer term.

cor3ntin:
@stephanemagnenat I'm already on it !

stephanemagnenat:
@car3ntin good, thank you. I am now taking the train and thus being online. Once the http is fixed, the challenge pack can be modified to use it instead of massloader. I let you merge the result, please do not forget to rebase so that the history is clean, thank you! have good holidays, I'll probably be available by email but only have phone and tablet.

davidjsherman:
I don't understand this comment.

cor3ntin:
@davidjsherman https://github.com/aseba-community/aseba/pull/765/files#diff-b2d2ad6e6d4199532251c463f6985340R20

`--port` ( or `-s` ) is not valid here.

1. Playground starts, launch its own server on `33333`, which the simulated thymios connect to
2. `asebahttp` is started by playground
3 `asebahttp` tries to start a server ( listening socket ) on that same `33333` port
4. nothing works.

removing `--port 33333` let `asebahttp` expose the playground robots through the http api without this issue.

cor3ntin:
however, I did indeed break the `--aesl` option of `asebahttp`

davidjsherman:
If the e2e tests didn't catch this, could you propose a test scenario that could have caught it?

PR #782, opened 2018-01-16
In Aseba compiler, num and bool values anywhere in expr and conditional statements

ypiguet-epfl:
The goal is to allow to store easily boolean values in variables and to use them in `if`/`when`/`while` without superfluous comparisons, and to give more flexibility in conditions. For example:
```
var a = prox.ground.reflected[0] > threshold
if a then
  ...
end
```
or
```
if (a > 5) ^ (b < 2) then
  ...
end
```
Both would require needlessly convoluted code with the current restrictions.

### DIFF
--- a/compiler/compiler.h
+++ b/compiler/compiler.h
@@ -4,16 +4,16 @@
 		Stephane Magnenat <stephane at magnenat dot net>
 		(http://stephane.magnenat.net)
 		and other contributors, see authors.txt for details
-	
+
 	This program is free software: you can redistribute it and/or modify
 	it under the terms of the GNU Lesser General Public License as published
 	by the Free Software Foundation, version 3 of the License.
-	
+
 	This program is distributed in the hope that it will be useful,
 	but WITHOUT ANY WARRANTY; without even the implied warranty of
 	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 	GNU Lesser General Public License for more details.
-	
+
 	You should have received a copy of the GNU Lesser General Public License
 	along with this program. If not, see <http://www.gnu.org/licenses/>.
 */
@@ -50,8 +50,8 @@ namespace Aseba
 	struct AssignmentNode;
 	struct TupleVectorNode;
 	struct MemoryVectorNode;
-	
-	//! A bytecode element 
+
+	//! A bytecode element
 	struct BytecodeElement
 	{
 		BytecodeElement() = default;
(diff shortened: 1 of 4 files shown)


stephanemagnenat:
Thanks for writing a PR with your idea, it is interesting.

As we discussed by email, I would be open to add support for boolean types explicitly through type inference, but not implicit casting from boolean to integers or the reverse, as this is a weak typing that is bad for education and a remnant of ancient languages such as C. My understanding of your PR at this stage is that it weakens the type system of AESL to allow implicit casting between booleans and integers. For example, this would allow:
```
var a = prox.ground.reflected[0] > threshold # we have boolean here
var b = 2 # we have integer here
var c = a == b # we compare a boolean and an integer, implicit casting
```
and also allow
```
var a = 0 == 0   # we have boolean here
if a == 0 then   # we interpret it as integer, implicit casting
    ...
end
```
Right?

Also, in your example ``if (a > 5) ^ (b < 2) then`` you are using an integer xor operator between booleans, while in Aseba logical operators between booleans are spelled out `and`, `or`, `not` (we could add `xor` as a convenience if deemed useful).

So I won't merge this PR until there is proper type inference and type checking between the types of variables. This is clearly a much larger work than the changes in this PR, but type safety is a critical concept especially in educational context and I won't compromise on it.

ypiguet-epfl:
Right, my request relaxes current type constraints. I must have misunderstood your email.

How would you write my first code example in a way you'd be ready to explain in an educational context?
As a more complete program for the Thymio, here is something I frequently see:
```
var running = 0
timer.period[0] = 10

onevent button.forward
  running = 1

onevent button.center
  running = 0
  
onevent timer0
  if running == 1 then
    // do some acquisition, computation and motion
  end
```
Same question, how would you write it?

cor3ntin:
How about
```
var running = false
timer.period[0] = 10

onevent button.forward
  running = true

onevent button.center
  running = false
  
onevent timer0
  if running  then
    // do some acquisition, computation and motion
  end
```

ypiguet-epfl:
cor3ntin, I like that, it's clean and doesn't require explanations. That's exactly what I seek. But it's impossible to compile it now, hence this PR. The next one would have been to predefine `false` and `true`.

cor3ntin:
I talked about it with @stephanemagnenat and we both agree that the correct approach is to have a `bool` type and no implicit casting ( or, at the very least, be very clear about what the casting rules are ).
We will work on adding a `bool` type to the language in the coming months, though we cannot promise a date at this time. That will also open the doors to more types in the future.

I understand that you have a need for that feature now, however merging your PR would prevent us from making the changes we want as we wouldn't be able to remain backward compatible.

In the meantime, you can use Aseba from your own branch. I realize it's not optimal.

ypiguet-epfl:
Ok, thanks for your comment.
I don't think there is one correct and many incorrect approaches, just different alternatives which attempt more or less successfully to provide a good balance between contradictory objectives. In a computer, everything is represented as numbers, so I don't think it's a problem to be forced to explain it even to young children. You cannot ask the compiler to catch all errors anyway. Aseba could easily catch more, overflows for instance.
Just by curiosity, would you accept explicit typecasts? Something like `int(2 < 3)` or `bool(1)`?

cor3ntin:
I think that "In a programming language there are types, integers can be used to do maths while Booleans can be true or false" is a better goal than "everything is a number" / "everything is whatever you need it to be", because then people grow up to invent JavaScript and we should avoid that.

There is a difference between statically checking the AST at parsing time and detecting overflows since the later is more of a runtime issue. And while we could do that, what should we do once an overflow is detected ? Beside  letting them happen and crashing the program there isn't much we can do.

I don't see why we couldn't have type conversion either int(), bool() or to_xxx()

ypiguet-epfl:
I won't argue here about the first paragraph, I'll just say I disagree. For runtime errors, out-of-bound array indices and divisions by 0 are already detected and signaled in Aseba Studio.

marvelous:
There is an issue with explicit casts as well, because the value mapping is implicit.
Why write `b = bool(x)` when `b = x != 0` is much more obvious? Bool to int conversion could benefit from an if expression: `speed = if running then 100 else 0`. There's no need to go as terse as C's ternary operator either.

cor3ntin:
@marvelous I actually think `b = x != 0` is probably the better option for teaching.

ypiguet-epfl:
marvelous, it was a rhetorical question. I'm not going to implement `bool()` and hope to have it accepted. But to answer you, we could also decide that false is 0 and true is 1, or -1, and document it.

marvelous:
That was my point. Either it's explicit in the program, or it's implicit and hopefully documented. I'm arguing for explicit.

stephanemagnenat:
I agree with @marvelous and also strongly vote for explicit.

ypiguet-epfl:
What was the question? Who's supposed to do what?
That's leading to nowhere. I'll stop here.

PR #784, opened 2018-01-16
Static stdlibcpp

cor3ntin:
the debian/rules should be something like

```
cd debian/build && cmake -D USER_BUILD_VERSION=${DEBVERS} -DASEBA_STATICLIBSTDC:BOOL=ON -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=../tmp/usr -DCMAKE_CXX_COMPILER=g++-5 -DCMAKE_C_COMPILER=gcc-5 ../..
```

Note sure you we could express that in a generic manner (aka only for the compiler when we have no choice)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -57,6 +57,14 @@ if (DEFINED CMAKE_SHARED_LINKER_FLAGS)
 	STRING(REPLACE "-Wl,--no-undefined" "" CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS}")
 endif (DEFINED CMAKE_SHARED_LINKER_FLAGS)
 
+
+if (UNIX)
+    if (ASEBA_STATICLIBSTDC)
+		set(CMAKE_EXE_LINKER_FLAGS "-static-libstdc++")
+		set(CMAKE_SHARED_LINKER_FLAGS "-static-libstdc++")
+	endif ()
+endif ()
+
 # testing and defines
 enable_testing()
 add_definitions(-Wall)
--- a/debian/control
+++ b/debian/control
@@ -3,7 +3,7 @@ Maintainer: Stephane Magnenat <stephane@nospam.magnenant.net>
 Section:devel
 Priority:optional
 Standards-Version: 3.7.2
-Build-Depends: cmake (>=2.6), libdashel (>=1.1.0), libenki (>=2.0~git.20180109), libqt4-dev, libqt4-opengl-dev, libqtwebkit-dev, qt4-dev-tools, libqwt5-qt4-dev, libudev-dev, libxml2-dev, libsdl2-dev, libavahi-compat-libdnssd-dev
+Build-Depends: cmake (>=2.6), libdashel (>=1.1.0), libenki (>=2.0~git.20180109), libqt4-dev, libqt4-opengl-dev, libqtwebkit-dev, qt4-dev-tools, libqwt5-qt4-dev, libudev-dev, libxml2-dev, libsdl2-dev, libavahi-compat-libdnssd-dev, g++ (>=5.0) | g++-5 | g++-7
 
 Package:aseba
 Architecture: any


cor3ntin:
D'autre part j'ai eu des soucis au moment de l'installation à cause de `-DCMAKE_INSTALL_PREFIX=../tmp/usr`. tu es sur de ça ?

CMakeLists.txt
```diff
@@ -57,6 +57,14 @@ if (DEFINED CMAKE_SHARED_LINKER_FLAGS)
 	STRING(REPLACE "-Wl,--no-undefined" "" CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS}")
 endif (DEFINED CMAKE_SHARED_LINKER_FLAGS)
 
+
+if(UNIX)
+    if(ASEBA_STATICLIBSTDC)
```
stephanemagnenat: indentation problem, please use tab

stephanemagnenat:
If I understand correctly, this should allow to compile on Trusty just by setting `ASEBA_STATICLIBSTDC` to true and forcing the g++ version? I am wondering, does the flag `ASEBA_STATICLIBSTDC` really belongs in the Aseba `CMakeLists.txt`? Shouldn't one instead directly specify `CMAKE_EXE_LINKER_FLAGS` and `CMAKE_SHARED_LINKER_FLAGS` for backports? I would like to avoid cluttering the main `CMakeLists.txt` file for very specific cases.

CMakeLists.txt
```diff
@@ -57,6 +57,14 @@ if (DEFINED CMAKE_SHARED_LINKER_FLAGS)
 	STRING(REPLACE "-Wl,--no-undefined" "" CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS}")
 endif (DEFINED CMAKE_SHARED_LINKER_FLAGS)
 
+
+if(UNIX)
```
stephanemagnenat: Normally the style is to have a space between the if and the (VAR), although there were inconsistent edits in the past.

cor3ntin:
It's necessary - albeit not sufficient to build on trusty. See my first comment.
I do thinks it's a valid thing to have in the cmake file as more distributions may have the issue.

stephanemagnenat:
> I do thinks it's a valid thing to have in the cmake file as more distributions may have the issue.

I see the point, I feel unease with hard-coding the name of a specific libstdC++ (for example on Android there are different ones).

cor3ntin:
Going forward we could either use another variable for libc++ or use the
same variable and select the correct c++ STD lib. In any case there is only
2 implementation of the STL and I don't think any distro uses libc++ by
default.

stephanemagnenat:
My point was whether, rather than adding explicit support for static c++ library in the CMakeList.txt, to simple do this on the command line: 
```
cd debian/build && cmake -D USER_BUILD_VERSION=${DEBVERS} -DCMAKE_EXE_LINKER_FLAGS="-static-libstdc++" -DCMAKE_SHARED_LINKER_FLAGS="-static-libstdc++" -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=../tmp/usr -DCMAKE_CXX_COMPILER=g++-5 -DCMAKE_C_COMPILER=gcc-5 ../..
```

I have a pretty bad experience with `CMakeLists.txt` that become too complex for special cases and become unmaintainable.

stephanemagnenat:
Btw, should we compile Dashel and Enki also with static c++ libraries?

cor3ntin:
We may do that for enki since it could use some apis not shipped in trusty.
I don't think it's the case though. I suggest you try to build aseba on the
ppa using these changes and we reevaluate that or later?

stephanemagnenat:
I'm working on preparing the package for the PPA new. But I'm confused:
- You said previously in a private discussion that the libstdc++ shipped by the default with Trusty should be sufficient for Aseba, did experimentation showed otherwise?
- The Debian control files specifies `g++ (>=5.0) | g++-5 | g++-7`, which g++ will this install? If it does not install version 5, Aseba will not build because version 5 is specified in the rules files.

cor3ntin:
As I think I mentioned earlier, I made no attempt to make the rules files
generic which we should. g++-5 is hard-coded for trusty. which is less than
ideal but will let you build on trusty. At some point we may want to write
a script that tries various version of g++ until it finds a suitable one.


2018-01-17 15:22 GMT+01:00 Stéphane Magnenat <notifications@github.com>:

stephanemagnenat:
Trusty builds are done, closing this PR.

PR #790, opened 2018-01-19
Compile on CI using Qt5

davidjsherman:
macOS: tell CMake where to find Qt5
Windows: tell CMake where to find Qt5
Windows: compile on node labeled "windows-qt5"

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -87,18 +87,20 @@ pipeline {
 						echo "macos_enki_DIR=${env.macos_enki_DIR}"
 						echo "macos_dashel_DIR=${env.macos_dashel_DIR}"
 						CMake([label: 'macos',
+							   getCmakeArgs: "-DCMAKE_PREFIX_PATH=/usr/local/opt/qt5",
 							   envs: [ "enki_DIR=${env.macos_enki_DIR}", "dashel_DIR=${env.macos_dashel_DIR}" ] ])
 						stash includes: 'dist/**', name: 'dist-aseba-macos'
 					}
 				}
 				stage("Compile on windows") {
 					agent {
-						label 'windows'
+						label 'windows-qt5'
 					}
 					steps {
 						sh 'git submodule update --init'
 						unstash 'dist-externals-windows'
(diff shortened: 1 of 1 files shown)


stephanemagnenat:
Could this be simplified after aa4c3512?

stephanemagnenat:
@cor3ntin you merged this PR but @davidjsherman did not answer my question. Did you discuss it with him? I want to avoid hacks being introduced when not necessary, we should not hastily merge PR that have outstanding unanswered comments. For this one we now have a `CMAKE_PREFIX_PATH` define that we do not know whether we need.

cor3ntin:
@stephanemagnenat I don't really see that as a hack. Maybe they could be moved to env variables or but in a yaml file.

I think the actual question is whether `Jenkinsfile` should be moved to https://github.com/aseba-community/aseba-continuous-integration altogether as even without those variables there are some assumptions made about the build system and therefore it's doubtful it could be used as-is by someone else.

stephanemagnenat:
@cor3ntin my question was that whether this path is useful after aa4c351? Shouldn't we expect to find Qt5 in the system path or be set through environment variables?

I think that the question of moving the Jenkins files to the CI repository is valid, but I will let @davidjsherman answer because he has the overview of what should be where regarding to CI.

davidjsherman:
1. @stephanemagnenat is correct about the process, his review was still open and it is his call whether to agree to changes or withdraw the review. I didn't have time to answer his question, in fact I only noticed it yesterday.
2. No. The philosophy of Jenkinsfile and Dockerfile in SCM means that they are **in the same SCM as the source code** and are tracked in the same history. It is fundamental.
[Ref 1](http://blogs.collab.net/devopsci/ten-best-practices-for-continuous-integration)
[Ref 2](https://github.com/jenkinsci/pipeline-examples/blob/master/docs/BEST_PRACTICES.md)
[Ref 3](http://bvajjala.github.io/blog/2014/04/22/best-practices-of-continuous-integration/)
[Ref 4](https://jenkins.io/doc/book/pipeline/jenkinsfile/)

PR #798, opened 2018-01-25
WIP: CMake refactoring

stephanemagnenat:
I create this PR not to merge for now, but to have a place to discuss the changes from the CMake refactoring.


CMakeModules/aseba_conf.cmake
```diff
@@ -0,0 +1,10 @@
+add_library(aseba_conf INTERFACE)
```
stephanemagnenat: I do not understand what is this library, could you please add documentation here?

CMakeModules/zeroconf.cmake
```diff
@@ -0,0 +1,27 @@
...
+	add_library(libavahi ${PROJECT_SOURCE_DIR}/third_party/avahi-compat-libdns_sd/compat.c)
+	target_link_libraries(libavahi PUBLIC ${AVAHI_CLIENT_LIBRARIES} ${AVAHI_COMMON_LIBRARIES})
+	target_compile_definitions(libavahi PUBLIC -DDNSSD_AVAHI)
+	set(HAS_AVAHI_SUPPORT ON)
+endif()
+
+
+if (DNSSD_FOUND)
+	set(HAS_ZEROCONF_SUPPORT ON)
+	add_library(zeroconf INTERFACE)
```
stephanemagnenat: Why is this library not prefixed by aseba?

android/thymiovpl/src/org/mobsya/thymiovpl/DashelSerial.java
```diff
@@ -1,150 +0,0 @@
...
-    {
-        return epout;
-    }
-
-    boolean CloseDevice()
-    {
-        connection.close(); // Linux take care of releasing everything else ...
-        return true;
-    }
-}
```
stephanemagnenat: I just realised that thymio vpl 2 might be using this file (and the rest of Dashel android), we should move it somewhere not delete it. I will check later.

marvelous: It's ok, VPL2 has a modified copy of this file: https://github.com/aseba-community/thymio-vpl2/blob/master/android/src/org/mobsya/thymiovpl/DashelSerial.java

davidjsherman:
This PR doesn't even build in CI (see [log](https://ci.inria.fr/inirobot/job/aseba-community/job/aseba/view/change-requests/job/PR-798/1/console)), I'm not looking any more closely for now.
```
hudson.plugins.git.GitException: Command "git merge 3b35de80d5fdd47592b1c01d57e1f4ef37c5e5ea" returned status code 1:
stdout: Auto-merging src/clients/studio/plugins/ThymioBlockly/blockly
Adding as src/clients/studio/plugins/ThymioBlockly/blockly~HEAD instead
Automatic merge failed; fix conflicts and then commit the result.
```

CMakeModules/aseba_conf.cmake
cor3ntin: It's a "virtual library", if you want. It has no source.
It however sets a number of properties that are used by its dependencies ( transitively )

In particular, it sets a few defines ( ASEBA_ASSERT, WIN32_LEAN_AND_MEAN ), include paths (src), as well as features and properties.

It's equivalent of setting global properties, albeit cleaner since you can write a target that do not "link" on aseba_conf if you want for some reason not to inherit its properties.

Makes sense ?

https://cmake.org/cmake/help/v3.10/manual/cmake-buildsystem.7.html#build-specification-and-usage-requirements

CMakeModules/zeroconf.cmake
cor3ntin: It's not related to aseba.
aka it's not asebazeroconf - which also exists.

when linking to `zeroconf` a target will automatically get the includes and libraries defined by `zeroconf` aka DNSSD_LIBRARIES, AVAHI_LIBRARIES, thread, support, etc.

it exists only to aggregate properties under a single name.

cor3ntin:
@davidjsherman this branch is not ready to be merged. However it's on top of master with no conflict.

android/thymiovpl/src/org/mobsya/thymiovpl/DashelSerial.java
stephanemagnenat: thanks @marvelous, so we can delete it here.

CMakeModules/aseba_conf.cmake
stephanemagnenat: > Makes sense ?

Yes, thanks for the explanation.

CMakeModules/zeroconf.cmake
stephanemagnenat: Ok, thanks for the explanation. Another question: couldn't these properties directly be attached to asebazeroconf?

cor3ntin: We could but I think it's cleaner that way, more modular.
Asebazeroconf depends on zeroconf and zeroconf happens to be a set of dnssd / avahi library.
It helps hiding the nastiness of find packages so that we can reason in terms of libraries where a library is a set of includes, defines, properties, compiler flags and actual library ( .a / .so ).

A library can either be private or public, which I haven't worked a lot on yet. That controls whether the properties get inherited transitively or not.

stephanemagnenat: Ok, that seems to make sense. Do you plan to document these design decisions?

cor3ntin: I can put a command in the CMakefile once its done, or maybe more comment in the issue ?
CMake has the ability to generate a graphviz .dot from the dependencies tree Maybe that's a good resource ?
You are probably more experienced with graphviz than me, I'm not sure how to make the graph reasonably good looking.

stephanemagnenat: Comments are always good, as is a dependency tree. But the key elements would be a description of the rationale behind the organization. Obviously you re-organize the tree with some concepts in mind, it would be helpful to explain them so that people can understand the CMake files easily. The current one are very unsophisticated, so it is not so much of a problem for people to understand them quickly, but with more sophisticated ones, a design documentation is necessary.

cor3ntin: Note that I'm merely trying to implement what I understand to be good "modern cmake" practices  namely
 - Describe everything as a dependency tree
 - Use as little compilers flags as possible, warning levels, optimization levels, sanitizers, etc should be handled in CTest files.

stephanemagnenat: This is already a start of a design documentation :-)

davidjsherman: Good choices are bubble graphs or chord diagrams, depending on the topology. DependencyWheel from D3 is a typical choice for chord diagrams.

stephanemagnenat:
@cor3ntin what is the rationale for moving the sources into `src/`? It breaks the maintainer scripts and the code of projects using Aseba as sub-module, such as VPL2.

cor3ntin:
@stephanemagnenat 

- I do not want non-source folders to be part of the include path ( aka, currently one can do `#include "debian/..." )
-  It's a standard layout for projects ( albeit we could also rename `src` to `aseba` )
-  I plan on adding a bunch of files and directories at the root for windows and osx packaging as well as a few build script files
-  you can also clone dashel and enki at the root 
 - There will be new source folders for the launcher and future projects

The maintainers script have been fixed, I'll make sure that VPL2 keeps working properly too.
Of course if you are strongly opposed to that I'll change.

CMakeModules/dashel.cmake
```diff
@@ -0,0 +1,9 @@
+if(EXISTS ${PROJECT_SOURCE_DIR}/dashel/CMakeLists.txt)
+	add_subdirectory(${PROJECT_SOURCE_DIR}/dashel)
+	message("Using dashel from ${PROJECT_SOURCE_DIR}/dashel")
```
stephanemagnenat: Please use capital d for dashel.

CMakeModules/zeroconf.cmake
```diff
@@ -0,0 +1,28 @@
+option(ENABLE_DISCOVERY "Enable device discovery ( zeroconf)" ON)
```
stephanemagnenat: extra space before zeroconf

stephanemagnenat:
The case is strong enough, ok for the new layout. Personally I slightly prefer using `aseba` instead of `src` so that we can have a `third_party` directory at the top-level, but I am open for discussion. Where did you plan to put third-party sources?

cor3ntin:
@stephanemagnenat I will rename src to aseba then. third_party will be at the top level.

stephanemagnenat:
Ok, that makes sense.

cor3ntin:
@davidjsherman is the ci running in this ? it doesn't appears to

cor3ntin:
it's alive ! https://ci.inria.fr/inirobot/job/corentin/job/aseba/job/cmake_3x/

PR #809, opened 2018-01-31
Guidelines for pull requests

davidjsherman:


stephanemagnenat:
@davidjsherman are these guidelines still WIP? My feeling was that after our last meeting, we could merge it, right?

cor3ntin:
Btw, it may be better to have and use a WIP tag rather than relying on the PR/commit name

stephanemagnenat:
For me a label is fine, @davidjsherman good for you or was there a specific reason to put WIP in the titel?

davidjsherman:
We might want to treat WIP PRs differently in CI. I know how to filter based on the PR name. If someone can establish how to filter on a discovered tag (the [documentation]() doesn't say how) then why not.

The comment said “it may be better” but didn't say why, so I didn't take it seriously.

davidjsherman:
Commit 9c8862620419965a4c789c6a3ad787662391afc9 adds a paragraph with guidelines for adding tests, including what we discussed about adding tests before a fix is found. This was the last thing in my notes, so I am removing WIP.

By the way, when I try to use the [`DISABLED` test property](https://cmake.org/cmake/help/v3.10/prop_test/DISABLED.html), the test is performed anyway. @cor3ntin have you ever succeeded in disabling a test?

Issue #624, opened 2017-05-05
Robot in playground: provide default behaviour


stephanemagnenat:
Playground should provide a way to specify default robot behaviors through an AESL source code. Practically, there are different ways of doing so:
1. Add the source code of a given node as part of the text content of the robot tag.
2. Add a `sourceCode` attribute that links to an AESL file.

Solution 2. is more work as it requires a full AESL parser in Playground, but it might be the occasion of factoring the different versions these parsers throughout the Aseba codebase.

stephanemagnenat:
As a temporary solution, this is currently done by have `asebamassloader` loading an aesl file.

mbonani:
is the massloader solution sufficient?

stephanemagnenat:
Using massloader is a ugly hack in my opinion, so I would like to keep this issue open, but maybe it does not need to be solved for 1.7. However, I suggest to wait for deciding that once we have a clear idea of what should go in 1.7.

davidjsherman:
The second suggestion is cleaner but seems inconsistent when there is more than one robot. An Aesl file defines a *network*, so by associating an Aesl file with each node, we implicitly define independent networks. It seems to me that a Playground with more than one node should almost always run a Switch, so that nodes that can physically interact can also communicate.

Furthermore, an Aesl file is already a way to associate programs with nodes, so providing a second mechanism in a Playground file is redundant. The responsibility for sending the program to the nodes could be the responsibility of the Switch. The current Http switch already has this ability and we can resolve this issue using the current command-line options.

A further improvement would be to explicitly declare networks in the Playground file. I think this would improve clarity. It can be implemented now by spawning an `asebahttp` process for each network, but could in the future be implemented by adding Switch to the Playground (to avoid separate processes). For example a Playground with two networks could look something like this:

```
<aseba-playground>
  <world w="300" h="300" />

  <robot type="thymio2" x="100" y="100" nodeId="1" name="Thymio 0" />
  <robot type="thymio2" x="150" y="150" nodeId="2" name="Thymio 1" />
  <robot type="thymio2" x="200" y="200" nodeId="3" name="Thymio 2" />

  <networks>
    <network id="1" name="Network 1" aeslFile="thymio-default.aesl" />
    <network id="2" name="Network 2">
      <event size="2" name="Clock"/>
      <constant value="0" name="False"/>
      <node nodeId="2" name="thymio-II">...</node>
      <node nodeId="3" name="thymio-II">...</node>
    </network>
  </networks>
[...]

PR #819, opened 2018-02-05
Playground: testing multi-node networks

davidjsherman:
This PR proposes end-to-end testing for multi-node networks within a Playground. It uses CMake to build testing scripts that refer to the correct executables. A test for a proposed extension to the playground format is included.

The testing scenario comprises five simulated Thymio-II robots partitioned into three networks.

1. A solo node with a program loaded from an Aesl file
2. A two-node network with separate programs loaded from an external Aesl file 
3. A two-node network with separate programs specified from an Aesl network whose definition is inlined in the Playground

The programs are simple beacons, that broadcast an *alive* message with the id of the node evry three seconds.

Playground `test-1-processes.playground` illustrates how to build three networks in a five-node playground, using `<process>` tags to start **asebaswitch**, **asebamassloader**, and **asebahttp** (for inlined Aesl programs).

Playground `test-2-networks.playground` illustrates a proposed extension to the XML definition, suggested in https://github.com/aseba-community/aseba/issues/624#issuecomment-362414283. Since this test isn't implemented, the CMakeLists.txt declares that this test is expected to fail.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -11,7 +11,7 @@ add_subdirectory(compiler)
 add_subdirectory(vm)
 add_subdirectory(simulator)
 #add_subdirectory(e2e-http)
-
+add_subdirectory(e2e-playground)
 
 # test asebahttp
 # FIXME: port to asebahttp2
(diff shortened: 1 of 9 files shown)


stephanemagnenat:
One question, are the tests supposed to work on Windows (in msys64)?

davidjsherman:
Yes, the tests ought to work anywhere.

stephanemagnenat:
Given the explanation of https://github.com/aseba-community/aseba/issues/624#issuecomment-362414283, what would you like me to do with this PR? Should we put it WIP until we finish the discussion in #624?

PR #833, opened 2018-02-11
Greek Translations

paraskevasleivadaros:
Added the Greek Translations as you requested! :)


stephanemagnenat:
Thank you very much! Which name/email would you like to see in the credits? (you can send me by email to avoid spam).

PR #836, opened 2018-02-13
turkish translation - done

onurkahveci:
Turkish translation for aseba studio is done and proofread.
Other files (of extra plugins etc.) will be done in few days.


onurkahveci:
What is the reason of the problem, error? So that I can fix it?

mbonani:
It seems that they was an error with the build system, it seems fine now. Maintainer has to control but he is in holiday

Issue #746, opened 2017-12-12
Unicode characters are not recognised as alphanumeric on some versions of macOS


mbonani:
aseba on windows (probably linux)  accept in the code accentuate character. 
on macosx it make error " indentifier must begin with _ or an alpha numeric character, found unicode character instead"
In the description of the language it is normally not permit. But generated code coming from blockly could have some. We should be consistent.

stephanemagnenat:
On my mac is works.

stephanemagnenat:
It seems some locales do not consider unicode characters to be alphanumeric. According to [cpp reference.com](http://en.cppreference.com/w/cpp/locale/isalpha), it could be forced. Maybe forcing the locale to "UTF-8" could help.

stephanemagnenat:
Do we move that to 1.6.1?

mbonani:
yes

ypiguet-epfl:
The error message has typos which make it difficult to find in the source code. It's _Identifiers must begin with _ or an alphanumeric character, found unicode character 0x?? instead_ where _0x??_ stands for the hexadecimal unicode code point for the Basic Multilingual Plane (up to U+FFFF), something else beyond. It's for error `ERROR_INVALID_IDENTIFIER` thrown by method `Aseba::Compiler::tokenize` in [aseba/compiler/lexer.cpp](https://github.com/aseba-community/aseba/blob/master/compiler/lexer.cpp).

Suggestion: write your own unicode handling code and don't depend on Qt, c++ standard libraries, the OS, user settings, or any short-term hack.

cor3ntin:
The underlying issue have been found, it's indeed because the tool rely on the underlying environment settings ( which it should not) and for some reason on OSX that environment may not be utf8.
Expect a fix in the coming weeks.

However, we should definitively not reinvent the wheel. Why locale support in the STL is somewhat lacking, it's top of the class in Qt, when used properly.

ypiguet-epfl:
What about tools which use the compiler without a GUI, such as the switch? In mine, when compiled on macOS, I get the same error:
```
./vmshell --dis --src -
var é
Error at Line: 1 Col: 5 : Identifiers must begin with _ or an alphanumeric character, found unicode character 0xe9 instead
Compilation error.
```
Unicode character categories and utf-8 aren't rocket science. A platform-independent, Qt-independent, definitive solution would be nice.

cor3ntin:
[...]

PR #842, opened 2018-02-24
Force utf8 locale while compiling

cor3ntin:


cor3ntin:
#746

cor3ntin:
There is a slight bug, I will be fixing that soon !

PR #852, opened 2018-03-08
Update the compilation instructions

cor3ntin:
See https://github.com/cor3ntin/aseba/blob/building_md/docs/en/development/building.rst


cor3ntin:
@mbonani can we merge that ?

docs/en/development/building.rst
```diff
@@ -0,0 +1,188 @@
...
+
+Getting Started on OSX
+----------------------
+
+You will need OSX 10.11 or greater
+
+-  Install `Homebrew <https://brew.sh/>`__.
+-  In the cloned repository run
+
+   brew update brew tap homebrew/bundle brew bundle
```
mbonani: this is a command it should be shown as the others

docs/en/development/building.rst
```diff
@@ -0,0 +1,188 @@
...
+~~~~~~~~~~~~~~
+
+In a msys2 prompt, run
+
+::
+
+    mkdir build && cd build
+    cmake -DBUILD_SHARED_LIBS=OFF -DCMAKE_BUILD_TYPE=Release -DCMAKE_PREFIX_PATH="<path of qt>;<path of bonjour>" ..
+    make
+
```
mbonani: this implies that the git clone was done inside msys2 no? perhaps better explain that somehow

cor3ntin: It does not, you just need to be in the location the repository was clone in. that may be anywhere on your drive

mbonani: ok

mbonani:
we can merge when the conflict is solved

PR #854, opened 2018-03-08
Do not copy headers in the installation folder

cor3ntin:
Aseba is more a set of tools than a framework.
We do not make ABI or API stability guarantees
except for some very specific classes.

Therefore we should not encourage people to use
aseba as libraries.

Modification, research works and new targets should rather
be added in-tree.

### DIFF
--- a/aseba/common/CMakeLists.txt
+++ b/aseba/common/CMakeLists.txt
@@ -17,30 +17,6 @@ install(TARGETS asebacommon
 	ARCHIVE DESTINATION ${LIB_INSTALL_DIR} 
 )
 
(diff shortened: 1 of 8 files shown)


davidjsherman:
>research works and new targets should rather be added in-tree

A product that uses Aseba is not necessarily part of Aseba. Needless strong coupling should be avoided

mbonani:
@davidjsherman I agree that a product or third party is not necessary part of aseba. Butr is there a sense to distribute all the lib and header file in the packet we provide?

cor3ntin:
@davidjsherman I applied the PR in Mobsya's fork as this was mostly relevant to Mobsya.

PR #864, opened 2018-03-14
Fix utf8 locale on windows

cor3ntin:
* fix utf8 locale on windows.


ypiguet-epfl:
Isn't the [boost license](http://www.boost.org/LICENSE_1_0.txt) missing, for [aseba/blob/master/third_party/catch2/include/catch2/catch.hpp](https://github.com/aseba-community/aseba/blob/master/third_party/catch2/include/catch2/catch.hpp)?

cor3ntin:
@ypiguet-epfl you are right, it would be better to add it even though I don't think it's strictly necessary. I have made a separate PR.

